PR #7, opened 2017-04-26
BCN-ISAK

isaklafleur:


views/artists.ejs
```diff
@@ -5,8 +5,10 @@
     <ul class="item-list">
       <% artists.forEach(function(artist) { %>
         <li class="single-item col-sm-4">
-          <% if(artist.images[0] != undefined) {%>
+          <% if(artist.images[0] != undefined) { %>
```
tawebbcn: This is a good way to solve a blank image, you could also create a function to control that and call it there, something like:

function imageCleaner(spotifyObject){
  if (spotifyObject.images[0]){
    return spotifyObject.images[0].url;
  } else {
    return "https://placeholdit.imgix.net/~text?txtsize=33&txt=250%C3%97250&w=250&h=250"
  }
}


And when you print the album you call the function to check if there is an image or not

app.js
```diff
@@ -65,3 +65,4 @@ app.get('/tracks/:albumId', (req, res) => {
 
 // start server
 app.listen(3000, () => console.log('The server has started'));
+
```
tawebbcn: good job!

PR #16, opened 2017-04-26
BCN-Giuseppe-Fierro

gfrr:
everything works, i have two issues though:
- when i search for an invalid artist i cannot handle the error, i tried few approaches but i didn't manage to solve the problem
- the artists and albums views are basically the same and a template would've been better, i'll try to fix this in the next commit


matilore:
In order to be able to pass invalid artists you just need to debug the response. If you check the response from spotify the artists are inside this array -> data.body.artists.items

so the best way to avoid errors is not to pass anything if that array is empty
So you could write something like this ->

app.get('/artists', (req, res, next)=>{
  console.log(req.query.artist);

  spotify.searchArtists(req.query.artist, {}, (err, data) => {
    if (err) throw err;
    if (data.body.artists.items.length != 0) {
      let artists = data.body.artists.items;
      res.render("artists", {artists: artists});
    } else {
      res.send("there are no artists with this name")
    }





  });

});

PR #21, opened 2017-07-05
BCN-julius-suominen

jsur:
done, no styling


tawebbcn:
Nice logic structure  🥇   really well organized

PR #24, opened 2017-07-06
bcn-denis-barriere

denisbarriere:
My credentials are in a folder /private ignored by git


tawebbcn:
The Token has an expiration time, doing that make for us too hard to check the exercise.
You need a post route to send the search to the api.

Api calls are async usually, take that in mind (because for example the spotify api sometimes is too slow). Check [this link ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)

PR #35, opened 2017-09-07
BCN - Lab-Express-Spotify: first and second iteration done

MHFP:


starter_code/express-spotify-app/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "express-spotify-app",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
tawebbcn: Try to add more scripts. At least one for starting the app.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,47 @@
...
+
+//SpotifyWebApi
+const SpotifyWebApi = require('spotify-web-api-node');
+const spotifyApi = new SpotifyWebApi({
+  clientId : '44010391e08c4ed88f1db7339ed9601b',
+  clientSecret : '34194b9295f8401f9905ea8719bf5862'
+});
+
+// Retrieve an access token.
+spotifyApi.clientCredentialsGrant()
```
tawebbcn: This piece of code is asynchronous, so you can have problems if the response is not ready when the app needs it. You can try to wrap it in a function with a callback.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,47 @@
...
+  .then(function(data) {
+    spotifyApi.setAccessToken(data.body.access_token);
```
tawebbcn: Are you sure this is how the spotify documentation says it's done?

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,47 @@
...
+
+app.use(express.static('public'));
+app.use(expressLayouts);
+
+app.get('/', (request, response, next) => response.render("index"));
+app.get('/artist', (req, res, next) => {
+    let artist = req.query.artist;
+    spotifyApi.searchArtists(artist).then((data) => {
+      let artists = data.body.artists.items;
+      res.render('artist', {artists});
```
tawebbcn: Be careful, you have named your view 'artists', not 'artist'.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,47 @@
...
+    }, (err) => {
+      console.log(err);
+    });
+});
+
+
```
tawebbcn: What about catching the errors and forwarding to the error handler?

starter_code/express-spotify-app/views/index.ejs
```diff
@@ -0,0 +1,11 @@
+
+
+
+  <h1>Search for an Artist</h1>
+
+  <form action="GET">
```
tawebbcn: Are you sure you want to do a GET? Not a POST or redirect to another route?

starter_code/express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,14 @@
+
+<section>
+
+    <% artists.forEach((artist) => { %>
+        <div>
+            <h1>Your search result:</h1>
+            <h2><%= artist.name %></h2>
+            <% if (artist.images.length > 0) { %>
```
tawebbcn: Good call! However, the best approach is define a helper function that, if there is an image, return it, otherwise displays a placeholder. Then you can use that function in albums.ejs too.

tawebbcn:
Be careful with asynchronous code and error handling. Anyways, good job! :+1:

PR #44, opened 2017-09-07
BCN - Philip Wisner

philipwisner:
It works. The preview only works on certain artists. I'm assuming that's just a copyright issue, not my code.

### DIFF
--- a/starter_code/express-spotify-app/app.js
+++ b/starter_code/express-spotify-app/app.js
@@ -46,25 +46,17 @@ app.get('/', (request, response, next) => {
 app.get('/artists', (request, response, next) => {
   spotifyApi.searchArtists(request.query.artists, {}, (err, data) => {
     if (err) throw err;
-    let artist =  data.body.artists.items;
-    response.render('artists',{ artist });
+    let artists =  data.body.artists.items;
+    response.render('artists',{ artists });
   });
 });
 
-app.get('/artists', (request, response, next) => {
-  spotifyApi.searchArtists(request.query.artists, {}, (err, data) => {
+app.get('/albums/:artistId', (request, response, next) => {
+  spotifyApi.searchArtists(request.query.artists.albums, {}, (err, data) => {
     if (err) throw err;
-    let artist =  data.body.artists.items;
+    let artist =  data.body.artists.albums.items;
     response.render('artists',{ artist });
   });
 });
 
-app.get('/albums/:artistId', (req, res) => {
-  spotifyApi.searchArtists(request.query.artists, {}, (err, data) => {
-    if (err) throw err;
-    let artist =  data.body.artists.items;
-    response.render('artists',{ artist });
-  });
-});
-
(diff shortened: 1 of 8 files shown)


starter_code/express-spotify-app/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "express-spotify-app",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
tawebbcn: Try to add more scripts. At least one for starting the app.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,71 @@
...
+const clientId = '8f1e69c9802e435e997e1983ec9348d0',
+    clientSecret = '4037307a11fd4f2ebc93931ea9b99e25';
+
+const spotifyApi = new SpotifyWebApi({
+  clientId : clientId,
+  clientSecret : clientSecret
+});
+
+// Retrieve an access token.
+spotifyApi.clientCredentialsGrant()
```
tawebbcn: This piece of code is asynchronous, so you can have problems if the response is not ready when the app needs it. You can try to wrap it in a function with a callback.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,71 @@
...
+  .then(function(data) {
+    let tracks = {
+      tracks: data.body.items
+    };
+    res.render('tracks', tracks);
+  }, function(err) {
+    console.error(err);
+  });
+});
+
```
tawebbcn: What about catching the errors and forwarding to the error handler?

starter_code/express-spotify-app/views/albums.ejs
```diff
@@ -0,0 +1,13 @@
+<h1>Albums Page</h1>
+
+<div>
+  <div>
+    <% for (let i = 0; i < albums.length; i++) { %>
```
tawebbcn: Better a forEach().

starter_code/express-spotify-app/views/albums.ejs
```diff
@@ -0,0 +1,13 @@
...
+      <h3><%= albums[i].name %></h3>
+        <% if (albums[i].images[0]) { %>
```
tawebbcn: Good call! However, the best approach is define a helper function that, if there is an image, return it, otherwise displays a placeholder. Then you can use that function in artists.ejs too.

starter_code/express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,13 @@
+<h1>Artist Page</h1>
```
tawebbcn: The same that with albums.ejs.

tawebbcn:
 - You are right, the error with the preview is not your fault.
 - What about some styles to make the app more beautiful?
 -  Amazing job with the commits!!!

Anyways, good job! :+1:

PR #45, opened 2017-09-07
BCN - Byron DE-Spotify

ByronGBP:


starter_code/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "starter_code",
+  "version": "1.0.0",
+  "description": "",
+  "main": "app.js",
+  "scripts": {
```
tawebbcn: Try to add more scripts. At least one for starting the app.

starter_code/app.js
```diff
@@ -0,0 +1,64 @@
+
+var SpotifyWebApi = require('spotify-web-api-node');
+var clientId = '2bb8357c009d414aa977da5341a6ac1a',
+    clientSecret = 'ef0d1a3f23a14ef8a5cd4934493969cd';
+var spotifyApi = new SpotifyWebApi({
+  clientId : clientId,
+  clientSecret : clientSecret
+});
+spotifyApi.clientCredentialsGrant()
```
tawebbcn: This piece of code is asynchronous, so you can have problems if the response is not ready when the app needs it. You can try to wrap it in a function with a callback.

starter_code/app.js
```diff
@@ -0,0 +1,64 @@
...
+    let album = req.params.album;
+    spotifyApi.getAlbumTracks(album).then((data) => {
+      let tracks = data.body.items;
+      console.log(tracks);
+      res.render('tracks', {tracks});
+    }, (err) => {
+      console.log(err);
+    });
+});
+
```
tawebbcn: What about catching the errors and forwarding to the error handler?

starter_code/views/album.ejs
```diff
@@ -0,0 +1,12 @@
+<a href="/"><button>BackHome</button></a>
+<div class>
+        <% albums.forEach((album) => { %>
+        <div>
+            <h3><%= album.name %></h3>
+            <% if (album.images.length > 0) { %>
```
tawebbcn: Good call! However, the best approach is define a helper function that, if there is an image, return it, otherwise displays a placeholder. Then you can use that function in artists.ejs too.

starter_code/views/album.ejs
```diff
@@ -0,0 +1,12 @@
+<a href="/"><button>BackHome</button></a>
```
tawebbcn: Maybe you can put this button in the layout and call it Home (instead of Back Home).

tawebbcn:
 - What about some styles to make the app more beautiful?
 - Just one commit? Seriously?!

Anyways, good job! :+1:

PR #46, opened 2017-09-08
BCN - Jennie DE 

JennieDalgren:


express-spotify-app/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "express-spotify-app",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
tawebbcn: Try to add more scripts. At least one for starting the app.

express-spotify-app/app.js
```diff
@@ -0,0 +1,73 @@
...
+app.set('views', __dirname + '/views');
+app.set('view engine', 'ejs');
+app.set('layout', 'layouts/main-layout');
+
+
+app.use(express.static('public'));
+app.use(expressLayouts);
+
+// Retrieve an access token.
+spotifyApi.clientCredentialsGrant()
```
tawebbcn: This piece of code is asynchronous, so you can have problems if the response is not ready when the app needs it. You can try to wrap it in a function with a callback.

express-spotify-app/app.js
```diff
@@ -0,0 +1,73 @@
...
+  .then((data) => {
+    res.render('tracks', {tracks: data.body.items} );
+  }).catch((err) => {
+    console.log(err);
+    throw err;
+  });
+});
+
+
+
```
tawebbcn: What about catching the errors and forwarding to the error handler?

express-spotify-app/views/albums.ejs
```diff
@@ -0,0 +1,16 @@
+<h1>Albums Page</h1>
+<div class="wrapperboxes">
+  <div>
+      <% for (let i = 0 ; i < albums.length; i++){ %>
```
tawebbcn: Better a forEach().

express-spotify-app/views/albums.ejs
```diff
@@ -0,0 +1,16 @@
...
+
+        <h3><%= albums[i].name %></h3>
+      <%  if (albums[i].images[0] ) {   %>
```
tawebbcn: Good call! However, the best approach is define a helper function that, if there is an image, return it, otherwise displays a placeholder. Then you can use that function in artists.ejs too.

express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,17 @@
+<h1>Artist Page</h1>
```
tawebbcn: The same that with albums.ejs.

tawebbcn:
 - Nice styles!
 - Amazing job with the commits!!!

Good job! :+1:

PR #48, opened 2017-09-08
bcn-jviles spotify initial work

jviles:


express-spotify-app/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "n",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
tawebbcn: Try to add more scripts. At least one for starting the app.

express-spotify-app/app.js
```diff
@@ -0,0 +1,18 @@
...
+var clientId = '61174fc8ca5c403a84ecbb20d47b131a',
+    clientSecret = '91f38e31ccd24dc18820b515bfcf28d1';
+
+var spotifyApi = new SpotifyWebApi({
+  clientId : clientId,
+  clientSecret : clientSecret
+});
+
+// Retrieve an access token.
+spotifyApi.clientCredentialsGrant()
```
tawebbcn: This piece of code is asynchronous, so you can have problems if the response is not ready when the app needs it. You can try to wrap it in a function with a callback.

tawebbcn:
Ya sabes que aún queda mucho trabajo por delante, pero por algo se empieza. :+1:

PR #50, opened 2017-09-27
PAR - Thibaut Davoult - Done

thibautdavoult:
...With super cool css from 1995


ta-web-paris:
Amazing. Actually useful for something done as a daily exercise!
Styling will be easier and less tedious later on with front end frameworks.

Keep it up

PR #51, opened 2017-09-27
PAR - Thomas Beguin - done

liloow:


liloow:
Sorry for class="...btn-danger..." I know I shouldn't use it just to make it red, but it's so much more convenient

ta-web-paris:
This is impressive! Great job on implementing the API calls and even nice views for a short exercise!

Don't worry too much about the "danger" class, styling will be easier in a real project using frameworks.

PR #52, opened 2017-09-27
PAR - Valentin - almost done

kira912:


ta-web-paris:
Bien joué :)
Good work on using the API. There is one issue with actually playing the tracks, I think it's a problem with using the audio controllers from the spotify API, you can try to go over it if you have time, but it's not so critical at the momemt, as long us you understood the concept.

PR #53, opened 2017-09-28
PAR - Jad Challita - done!

jadch:
Done with the application logic and styling of the landing page. Didn't do the rest of the styles.


ta-web-paris:
There's an issue with the spotify API token. Are you sure you're using the right ones? Otherwise, it could be a problem with the API... no worries

jadch:
I revoked/changed my client key after I noticed I uploaded it. Didn't know you would (or even could) use it, sorry!

jadch:
new one => 7d63e0d3baa04e1fa2b001110b426059

PR #54, opened 2017-09-28
PAR - Frederic Magnin

frederic117:
not finish at all


frederic117:
lot of help from maxence, stuart thomas, thibault valentin...

ta-web-paris:
Nice. The only thing not working yet is the actual specfic album view. I think you understand the concept of working with external APIs though :)
If you have any questions let us know.

PR #55, opened 2017-09-28
PAR - Nicolas - Iteration1

nicolasBarthez:
Je me suis arrêté à l'affichage des albums, je ne comprends pas pourquoi je n'arrive pas à afficher l'image... car mon console log m'affiche bien l'URL voulue...
Merci pour votre aide.

### DIFF
--- a/starter_code/express-spotify-app/views/albums.ejs
+++ b/starter_code/express-spotify-app/views/albums.ejs
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Search for an Artist on Spotify</title>
+    <link rel="stylesheet" type="css/text" href="../public/stylesheets/style.css">
+  </head>
+  <body>
+    <% albums.forEach(alb => { %>
+      <img src="
+      <%alb.images[0].url%>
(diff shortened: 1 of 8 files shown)


starter_code/express-spotify-app/views/albums.ejs
```diff
@@ -0,0 +1,17 @@
...
+    <title>Search for an Artist on Spotify</title>
+    <link rel="stylesheet" type="css/text" href="../public/stylesheets/style.css">
+  </head>
+  <body>
+    <% albums.forEach(alb => { %>
+      <img src="
+      <%alb.images[0].url%>
+      " alt="">
+      <p><%=alb.name%></p>
+      <a href="/tracks/<%=alb.id%>">View Tracks</a>
```
mlande: This link is routing the user to "/tracks/:id" route, but you don't have a handler for that route, so it is not showing anything. You will need to add an app.get route handler for that, and make it render a view.

mlande:
Nice work until the problem you mentioned :)
See my comment on the code. You are sending the user to a route you are not handling.
Let me know if you want to discuss this in person

PR #75, opened 2017-11-08
BCN - Clara Ameller

ClaraAmeller:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,79 @@
...
+    let data = {
+      track: tracks,
+    };
+    res.render("tracks", data);  })
+   .catch(err => {
+     console.log(`Ooops something went wrong while searching the albums -> ${err}`);
+  });
+});
+
+
```
```
app.use((req, res, next) => {
  err.status = 404;
  next(err);
});

// error handler

  // render the error page
  res.render('error');
});
```

starter_code/express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
```
tawebbcn: Maybe you can use a main-layout for all the pages.

PR #80, opened 2017-11-08
BCN - Torgeir Beyer

torgeirbeyer:
showing picture


starter_code/express-spotify-app/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "express-spotify-app",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
tawebbcn: Remember to add the necessary scripts.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,77 @@
...
+  //     console.log(artistAlbums);
+  //     // res.send(albums);
+  //   },
+  //   function(err) {
+  //     console.log(err);
+  //   }
+  // );
+  // res.render("view-albums");
+});
+
```
```
app.use((req, res, next) => {
  err.status = 404;
  next(err);
});

// error handler

  // render the error page
  res.render('error');
});
```

starter_code/express-spotify-app/views/layout/main-layouts.ejs
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
```
tawebbcn: Good call using a layout.

starter_code/express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,10 @@
+<h1>
+  <%= body.artists.items[0].name %>
```
tawebbcn: You are only showing the first result in the query, you need to use a for, forEach... If you did that to avoid the error of not being able to show url of undefined (some results doesn't have images), that doesn't solve the problem (try to search Japanese Beatles) and it's not the desired behaviour. You can solve this problem (and similar ones) with:
```
<% if(artists[i].images[0]){%>
  <img src="<%=artists[i].images[0].url%>" alt="" class="artist-img">
<%} %>
```

PR #83, opened 2017-11-09
hey how are you

MythicalMichael:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,66 @@
...
+
+
+})
+
+app.post("/artists", (req, res) => {
+    let artists = req.body.artists;
+    spotifyApi.searchArtists(artists).then(function(data){
+
+
+        res.render("artists",data);
```
tawebbcn: Better pass an object to render (e.g. {data} instead of data).

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+   );
+
+  });
+
+
+
+
+// -- 404
```
tawebbcn: Good comment, but don't forget the 404 and the error handler:
```
app.use((req, res, next) => {
  err.status = 404;
  next(err);
});

// error handler

  // render the error page
  res.render('error');
});
```

PR #85, opened 2017-11-09
BCN - Lucas

lacazeto:


starter_code/app.js
```diff
@@ -0,0 +1,51 @@
...
+      res.render('artists', data);
+    }, function(err) {
+      console.error(err);
+    });
+});
+
+app.get('/albums/:artistId', (req, res) => {
+    // code
+});
+
```
tawebbcn: Don't forget the 404 and the error handler:
```
app.use((req, res, next) => {
  err.status = 404;
  next(err);
});

// error handler

  // render the error page
  res.render('error');
});
```

starter_code/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "starter_code",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
tawebbcn: Remember to add the necessary scripts.

starter_code/views/artists.ejs
```diff
@@ -0,0 +1,18 @@
...
+<head>
+    <meta charset="utf-8">
+    <title>My first view</title>
+    <link rel="stylesheet" href="stylesheets/style.css">
+</head>
+
+<body>
+
+</body>
+    <h1><%= body.artists.items[0].name %></h1>
```
tawebbcn: You are only showing the first result in the query, you need to use a for, forEach... If you did that to avoid the error of not being able to show url of undefined (some results doesn't have images), that doesn't solve the problem (try to search Japanese Beatles) and it's not the desired behaviour. You can solve this problem (and similar ones) with:
```
<% if(artists[i].images[0]){%>
  <img src="<%=artists[i].images[0].url%>" alt="" class="artist-img">
<%} %>
```

PR #92, opened 2017-11-09
BCN - Eloi

eloigil:
Hi TAs!


starter_code/express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,8 @@
+<body>
+    <div class="container">
+
+    </div>
+    </div>
+</body>
+
+</html>
```
tawebbcn: If you are using a main-layout, you don't need the body tags, neither the html one.

starter_code/express-spotify-app/views/index.ejs
```diff
@@ -0,0 +1,12 @@
...
+        <div class="inputContainer">
+            <input type="text" id="artist" />
+        </div>
+        <div class="buttonContainer">
+            <a href=/artists id="submit">Submit</buton>
+        </div>
+    </div>
+</body>
+
+</html>
```
tawebbcn: Same as before, not body, not html.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,60 @@
...
+app.use(express.static('public'));
+app.use(expressLayouts);
+app.set('layout', 'layouts/main-layout');
+app.set('views', __dirname + '/views');
+app.set('view engine', 'ejs');
+
+// first Route
+app.get('/', (req, res) => {
+    console.log(req);
+    res.render('index.ejs');
```
tawebbcn: Don't put the extension:
```
res.render('index');
```

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,60 @@
...
+});
+
+//artists
+app.get('/artists', (req, res) => {
+    spotifyApi.searchArtists().then((result) => {
```
tawebbcn: You need to pass the artist to the method, e.g.:
```
spotifyApi.searchArtists(req.query.artist).then //....
```

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,60 @@
...
+        })
+        .catch((err) => {
+            console.log(err);
+        });
+});
+
+
+
+//listen
+
```
```
app.use((req, res, next) => {
  err.status = 404;
  next(err);
});

// error handler

  // render the error page
  res.render('error');
});
```

PR #96, opened 2017-12-27
PAR - done - @VivianSolide

VivianSolide:


ta-web-paris:
First thing, really good job! You understand how to create routes, call an API asynchronously and display views. 

I think the main problem you faced that you didn't solve is in EJS when a variable doesn't have a property and it shows an error. For that, you only need to add a condition to check if the property is defined.

I have written all the possible improvements you could have done for this exercise:

--------

You didn't respect the **iteration 3**, were you should display many result of the search. I think your goal was to avoid the next error (but I found a case were it doesn't work).

--------

I found an error at http://localhost:3000/artist?artist=le%20peuple : 
> ` >`
> **Cannot read property 'url' of undefined**

With EJS, you should always check if something is defined. The cleanest way would be the following code (yes I know, it's long and verbose):
```
<% if(songs[i] && songs[i].images && songs[i].images[0] && songs[i].images[0].url) { %>
 <img src= <%= songs[i].images[0].url %> >
<% } %>
```

-------

Another "error": when something bad happend (like at http://localhost:3000/artist), some errors are displayed on the terminal but nothing is displayed on the web page (it waits to render something). We are going to see an easy to solve that but in your case you could have written something like this: `res.render('error')`

---------

On the the album page (http://localhost:3000/albums/?id=3WrFJ7ztbogyGnTHbHJFl2), you didn't displayed any picture. Try to display a picture when there are some.

------------------

You am not sure you understood the "static/public" concept. In `app.js`, instead of 
```js
app.get('/public', (req, res, next) => {
  res.render('') // = res.render('index')
})
```

You should write:
```js
// Line 5
app.use(express.static('public'));
```

And now, for example, you have access to `public/stylesheets/style.css`with the link: http://localhost:3000/stylesheets/style.css

-------------------

In all you EJS files, you must change `/public/the/link/to/your/static.file` by `/the/link/to/your/static.file`.

You should have seen this error by opening the Chrome terminal.

---------

It could be nice to use template in this exercise (http://learn.ironhack.com/#/learning_unit/2413). It would avoid you putting HTML in all EJS files.

---------

In **package.json**, you could add the following line in your `scripts`: `"start": "node app.js"`

-------------------

I hope this feedback will help you!

Your master,
Maxence

VivianSolide:
Amazing feedback!
Indeed I didn't understand the _express/static_ module.
Thank you for your time!

PR #115, opened 2018-01-31
BCN - Rodrigo

rodrigodaoud:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,49 @@
...
+//index route
+app.get('/', (req, res, next) => {
+    res.render("index");
+});
+
+//artist route
+app.get('/artists', (req, res, next) => {
+    let artistSearch = req.query.artist;
+    spotifyApi.searchArtist(artistSearch).then((data) => {
+        res.render("artist", {data: body.artist.items});
```
tawebbcn: body is a property of your response (data), so you are missing that part:
```
res.render("artist", {data: data.body.artist.items});
```
Maybe is not a very good idea call data your response a the key of your object (that said, it will work).
Of course, you need a view called artist.

PR #119, opened 2018-01-31
BCN - LUIS

luisacherrera:


starter_code/express-spotify-app/views/index.ejs
```diff
@@ -0,0 +1,5 @@
+<body>
```
tawebbcn: Don't use body tag or you will have two bodys one inside the other (the one in layout and this one).

PR #123, opened 2018-02-01
BCN- DANI 

DTDK:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,43 @@
...
+  res.render('index', data); // I dont understand why we use response, is it a response
+});                          //  to the response of my request??
+
+
+app.listen(3000,() => {
+  console.log('Listeninng on port 3000');
+});
+
+// Remember to paste here your credentials
+var clientId = 'de05d80b9f2540d4a0d51f5344718aee',
```
tawebbcn: The code from here until the end has to be before your routes (line 14) so you can use spotifyApi in them.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,43 @@
...
+
+const app = express();
+
+
+app.set('views', __dirname + '/views');
+app.set('view engine', 'ejs');
+
+
+app.get('/', (req, res, next) => { 
+  let data = {
```
tawebbcn: Why do you do this? Does index need the strings "Spotify" and "application"?

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,43 @@
...
+    name: "Spotify",
+    type: "application"
+  };
+  res.render('index', data); // I dont understand why we use response, is it a response
```
tawebbcn: It is the response that we are sending to the browser.

PR #124, opened 2018-02-01
BCN - Linda 

lindaluisa:
Thanks for feedback, pce


express-spotify-app/public/javascripts/app.js
```diff
@@ -0,0 +1,52 @@
...
+app.set('views', __dirname + '/views');
+app.set('view engine', 'ejs');
+
+// -- ROUTE
+app.get('/', (req, res, next) => {
+  console.log("test " + req.method, req.path, req.query);
+  res.render('search', {title: "Hey", message: "Hello, there"});
+});
+
+// RETRIEVE AN ACCESS TOKEN
```
tawebbcn: For a better structure, put this before any route.

express-spotify-app/public/javascripts/app.js
```diff
@@ -0,0 +1,52 @@
...
+    console.log('Something went wrong when retrieving an access token', err);
+});
+
+
+// SEARCH ARTIST
+app.get('/search-for-artist', (res,req,next) => {
+  res.render('search')
+});
+
+app.get('/search-for-artist', (res,req,next) => {
```
tawebbcn: You can't have two routes with the same verb (get in this case) and the same path. The second one never will be reached.

PR #127, opened 2018-02-01
BCN Alla

allarom:
Start

### DIFF
--- a/starter_code/express-spotify-app/views/artists.ejs
+++ b/starter_code/express-spotify-app/views/artists.ejs
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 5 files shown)


starter_code/express-spotify-app/views/artists.ejs
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+    <title>My first view</title>
+  <link rel="stylesheet" href="../../public/stylesheets/main.css">
+</head>
+<body>
+  <h1>Spotify Login</h1>
+  <form action="/artists" method="post">
```
tawebbcn: You have this form in index.ejs. Use this page to display the results.

PR #128, opened 2018-02-01
BCN - Vlad 

vladcio:


starter_code/express-spotify-app/views/layouts/artists.ejs
```diff
@@ -0,0 +1 @@
+<p>artist</p>
```
tawebbcn: In the layout folder you only want your layout, don't put there another views.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,57 @@
...
+    spotifyApi.setAccessToken(data.body['access_token']);
+  }, function (err) {
+    console.log('Something went wrong when retrieving an access token', err);
+  });
+
+app.get('/', (req, res) => {
+  res.render('index', {message: null});
+});
+
+app.post('/artists', (req, res) => {
```
tawebbcn: This route is a get, you use it to display your artists page.

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,57 @@
...
+//   let artistSearched = req.body.artist;
```
tawebbcn: Somewhere in this route you need to use the searchArtists method from spotifyApi.

PR #131, opened 2018-02-01
BCN - Sheng

theRealSheng:
I need to review it further before able to finish it...

### DIFF
--- a/starter_code/express-spotify-app/app.js
+++ b/starter_code/express-spotify-app/app.js
@@ -1,0 +1,67 @@
+'use strict';
+
+const express = require('express');
+const app = express();
+const bodyParser = require('body-parser');
+const expressLayout = ('express-ejs-layouts')
+const morgan = ('morgan');
+const prettyJson = ('prettyjson');
+const SpotifyWebApi = require('spotify-web-api-node');
+
+app.set('views', __dirname + '/views');
+app.set('view engine', 'ejs');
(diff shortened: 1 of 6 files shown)


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,67 @@
...
+    res.render('home');
+});
+
+app.post('/artist', (req, res) => {
+
+    let artistSearched = req.body.artist;
+
+    spotifyApi.searchArtists(artistSearched)
+        .then((data) => {
+        res.render('artist', {data: body.artist.items});
```
tawebbcn: better create an object instead of passing it inside curly braces.

```
 let artistsData = {
  data : data.body.artists.items;
}
res.render('artist', artistsData);
```

starter_code/express-spotify-app/views/layout/layout.ejs
```diff
@@ -0,0 +1,15 @@
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Document</title>
+</head>
+<body>
+    <h1>Hello world</h1>
+    <div class="artist-wrapper">
```
tawebbcn: this is a wrapper for everything, not just for the artist

PR #132, opened 2018-02-01
PAR - Arthur - Spotify 

dudydude:
I had trouble to display the albums, I'm stuck with a [object [Object]], couldn't find solution..


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,51 @@
...
+  clientId: clientId,
+  clientSecret: clientSecret
+});
+
+app.get("/", (req, res, next) => {
+  res.render("index");
+});
+
+app.get("/artists", (req, res, next) => {
+  let artist = req.query.artist;
```
ta-web-paris: `req.query` is trying to fetch the query part of the url. It means the url should have some kind of query structure, with a `?`. Check out these docs http://expressjs.com/th/api.html, in the section *req.query*

starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,51 @@
...
+
+  spotifyApi
+    .searchArtists(artist)
+    .then(artist => {
```
ta-web-paris: I would name the callback argument something else that `artist` to prevent confusion with the `artist` variable you defined a few lines up

dudydude:
@ta-web-paris I made a new pull request with the code that return [object Object]

PR #134, opened 2018-02-01
BCN Stephie

kissy89:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,62 @@
...
+app.get('/', (req, res, next) => {
+    res.render('start');
+  });
+
+// search for an artist - spotifyAPI fct
+
+app.get("/artists", (req, res, next) => {
+    spotifyApi.searchArtists(req.query.artists)
+        .then((data) => {
+          var artists = data.body.artists.items;
```
tawebbcn: let

tawebbcn: better create an object instead of passing it inside curly braces.

```
 let artistsData = {
  data : data.body.artists.items;
}
```

PR #137, opened 2018-02-01
BCN Martin

martinX308:


martinX308:
I have a question about the css. what path do I have to enter here?   
<link rel="stylesheet" href="/public/stylesheetsa/style.css"> 
It seems not to be found

PR #182, opened 2018-04-12
[MAD DE] Luis Nueda - Week4 Day4

luisnueda:
1-Número de iteraciones: completado.
	2-Dificultad: Media-Alta
	3-Comentario: falta por asentar el get y post, me cuesta asimilar el uso de las apis, y el get y post o tengo que trabajar mas.


ta-web-mad:
#closed #checked #√ictor Luis apologize for the delay in the response. I do not think you have any doubts right now, because the weekend has already passed and you have settled the knowledge. In my opinion, it always clarifies the code to put the get first and then the post. By convention, it makes it clearer that I will get in that view and that I will send it to that view. But to be able to cook you have to buy, so GET and then POST. Sorry again.

PR #189, opened 2018-04-13
bcn - matias

Matuki92:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,56 @@
...
+    console.log('Something went wrong when retrieving an access token', err);
+  });
+
+// Routes
+app.get('/', (req, res, next) => {
+  res.render('search');
+});
+
+app.get('/artists', (req, res, next) => {
+  spotifyApi.searchArtists(req.query)
```
tawebbcn: `req.query` is an object that have the query that have the query that are in the path of the request. 
So you should specified what you want.

starter_code/express-spotify-app/views/search.hbs
```diff
@@ -0,0 +1,11 @@
+<form method="get" action="/artists">
+
+  <div>
+    <label></label>
+    <input type="text" name="artist">
+  </div>
+
+  <div>
+    <button type="submit">Search for an Artist.</button>
```
tawebbcn: `type="submit"` doesn't do anything if the button is not inside a form

PR #201, opened 2018-04-13
PAR Bertrand

wavilo:
sorry No CSS but the rest is working

### DIFF
--- a/starter_code/express-spotify-app/app.js
+++ b/starter_code/express-spotify-app/app.js
@@ -1,0 +1,94 @@
+var SpotifyWebApi = require("spotify-web-api-node");
+
+// Remember to paste here your credentials
+var clientId = "af3c2403fc0a4284aa9ea8a30724d5cd",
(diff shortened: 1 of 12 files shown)


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,94 @@
...
+//route albums
+app.get('/albums/:artistId', (req, res, next) => {
+
+  spotifyApi.getArtistAlbums(req.params.artistId)
+  .then(function(data) {
+    res.render("albums", {
+      data: data
+    });
+    // console.log('Artist albums', data.body.items);
+  }, function(err) {
```
ta-web-paris: If you are using a `.then` callback, the error handling should be in a `.catch` callback, and not as a second argument to the `then` callback. For example:
` spotifyApi.getArtistAlbums(req.params.artistId).then(function(data) {
 //doStuff 
}.catch(function(err) { 
//handleError
}
`

PR #203, opened 2018-04-13
PAR Renaud

Ayss83:


starter_code/express-spotify-app/app.js
```diff
@@ -0,0 +1,74 @@
...
+    .catch(err => {
+      console.log("Error", err)
+    });
+});
+
+app.get('/albums/:artistId', (req, res) => {
+  spotifyApi.getArtistAlbums(req.params.artistId, {limit: 10})
+  .then(data => {
+    // res.send(data);
+    res.render("albums", {
```
ta-web-paris: If the API returns to you already the albums, you could name the callback argument `albums` and then in the render function simply write: 
`res.render( "albums", { albums } )`
:)

starter_code/express-spotify-app/views/artists.hbs
```diff
@@ -0,0 +1,14 @@
...
+<div class="artist-card">
+  {{#each this.images}}
+    {{#if @first}}
+      <div class="albumCover">
+        <img src="{{this.url}}" alt="cover">
+      </div>
+    {{/if}}
+  {{/each}}
+  <h2>{{this.name}}</h2>
+  <a class="btn btn-danger" href="/albums/{{this.id}}">Show albums</a>
```
ta-web-paris: I get that you liked the `danger` color, but it's a better practice to use semantic classes, so use `danger` for something that is actually a danger for the user. You could always customize the colors :)
Not a big deal, but just giving some extra tips since you are doing well with the basics

PR #211, opened 2018-04-13
PAR Mostafa

M-Gadd:


ta-web-paris:
Good job!
Use the ```html<table>``` in [tracks.hbs](https://github.com/ironhack-labs/lab-express-spotify/pull/211/files#diff-213507b56448cd0a45e52f26c7f6beaa
) instead.
:tada: 
_vivian_

PR #212, opened 2018-04-13
PAR Sami

Sami-Az:
Still working on the albums page and further


ta-web-paris:
We will do the exercice together.
_Vivian_

PR #213, opened 2018-04-13
PAR Patrycja

kucharskap:
I struggled a lot and managed to finish only the first 3 iterations. I'll work on it this weekend to get a deeper understanding of the whole subject


ta-web-paris:
We will do the exercice together.
Probably during a night shift.
_Vivian_

PR #215, opened 2018-04-13
PAR: Leo P.

LPsola:
Haven't finished but took the time to get the logic down. will most likely finish it up this week end.


ta-web-paris:
It works! Good job.
You missed just this:
```js
<audio controls src="{{this.preview_url}}" ></audio>
```
_Vivian_

PR #217, opened 2018-04-13
PAR Pierre

PierreJab:


ta-web-paris:
:tada: 
Nice one!
Your app.js is really good.
You can improve your code by using partial.
DRY: Don't Repeat Yourself 
_Vivian_

PR #240, opened 2018-06-24
PAR Meri

mpligioso:
I have the feeling I did something weird when installing the packages (therefore two .gitignores as a precaution)... Also the styling is not really done done.

### DIFF
--- a/starter_code/.gitignore
+++ b/starter_code/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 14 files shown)


ta-web-paris:
Actually, when you clone a repo from GitHub it makes your local folder a repository, and when you generate an express project with Irongenerate it makes a repo too. So you could have deleted one of them yes, but that's not a big issue.

Otherwise, the logic is very well understood!
And the styling is well done too, even if that's not the most important :)

Continue comme ça !

PR #268, opened 2018-07-05
BCN - Montse Badia

MontseBadia:
10% done :(
Impossible to get the access to the API


starter_code/app.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const express = require('express');
+const app = express();
+const hbs = require('hbs');
```
tawebbcn: We requiere hbs but we are not using it, should be deleted that line of code.

starter_code/app.js
```diff
@@ -0,0 +1,65 @@
...
+  res.render('index');
+});
+
+app.get('/artists', (req, res, next) => {
+  // console.log(req.query);
+
+  spotifyApi.searchArtists(req.query.artist)
+  .then((data) => {
+    console.log(data.body.artists.items[0].images)
+    res.render('artists', {artists: data.body.artists.items})
```
tawebbcn: another way would be : ``` res.render('artist', data) ```
What we do it's render all the object and then in the view we can acces to what we want to show

tawebbcn:
Check the code comments but for the rest, very good Montse 👍

PR #289, opened 2018-07-06
BCN - Will

dambusm:


starter_code/app.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+const express = require('express');
+const app = express();
+const hbs = require('hbs');
```
tawebbcn: We're not using that variable so we shouldn't requiere it.

starter_code/app.js
```diff
@@ -0,0 +1,77 @@
...
+    .then(function (data) {
+      console.log(data.body);
+      res.render('tracks', data.body);
+    }, function (err) {
+      console.log('Something went wrong!', err);
+    });
+});
+
+// Start listening
+app.listen(3000);
```
tawebbcn: Instead of app.listen, it's better to launch with the script on the package.json.

tawebbcn:
Try to check the two comments and for the rest it's everythuing perfect,
Very good Will 👍

PR #320, opened 2018-09-03
PAR Abi & Julie

juliemnrd:
@abisummers


ta-web-paris:
This works fine!

The next steps would have required very similar mechanics, but with another spotifyAPI method: instead of using .searchArtists, you would have used the "search by id" one and refined the data the same way. Don't hesitate to do the exercise again and try to finish it when you have some spare time, practice makes perfect :)

Lucas

PR #321, opened 2018-09-03
john and emmanuelle

johndimitroff:
@emmanuellejaldorau

### DIFF
--- a/starter_code/express-spotify-app/app.js
+++ b/starter_code/express-spotify-app/app.js
@@ -1,0 +1,65 @@
+const express = require('express');
+const app = express();
+const hbs = require('hbs');
+const path    = require("path");
+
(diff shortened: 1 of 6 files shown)


ta-web-paris:
This works fine!

The `artists/:artistId`step was almost completed :)
Three things to get it right, and be able to push forward:
you "receive" an artist ID in your url, but
1.  before being able to use it you'll need to use req.params in some way. Check the lesson again to see how to do that!
2. when searching for artist albums, you gave the method a string. This won't work :)
3. be careful to always have your variables match! In the route you declare `res`, but use something else when trying to get some json data...

Don't hesitate to do the exercise again and try to finish it when you have some spare time, practice makes perfect ;)

Lucas

PR #327, opened 2018-09-06
MIA FTWD Sat

satga:
works but zero styles


MarcosPalaciosIII:
Excellent work on the express spotify clone. Good work adding a nav bar to link back to home page.

Keep up the excellent work.

PR #328, opened 2018-09-07
MIA FTWD alex rojas

alexrojas:


MarcosPalaciosIII:
Great job on the express spotify lab. It would be helpful if you would have added a nav bar to link back to home. 

Keep up the great work.

PR #329, opened 2018-09-07
MIA FTWD Bartashevich-Kiryl

bartosian:


MarcosPalaciosIII:
Excellent work on the express spotify lab. It would be nice to have a nav bar that would send you to the homepage incase someone wanted to search for another artist. 

Keep up the excellent work.

PR #430, opened 2018-11-21
AMS Dea

Deakristina:


IronhackAMS:
Hi Dea!

Creative solution putting the background img inline in hbs!:) Also nice that you put all the HTTP requests together in `app.use("/artist", routesSearch)`. Good Job!

PR #431, opened 2018-11-21
AMS Ana Kelly

anakellyc:


IronhackAMS:
Hey Ana!,

Well done! 
A couple of small remarks: 
- the partials folder is mostly used for partials like navbar, footers, or error handlers etc
- tip: error-handling npm is used within dev environment, if you'd like sent messages to client/user then connect-flash is a good option.

PR #432, opened 2018-11-21
AMS Matt Hamers 

mattviy:
finalized spotify API exercise


IronhackAMS:
Nice, very tidy code! Try to avoid using inline styling though.

PR #443, opened 2018-11-23
AMS David Bros

TheHunter896:
All functionalities required plus the web will tell you in case the album you have chosen doesn't have the previews avaliable


IronhackAMS:
Looks very good! See if you can omit requiring modules where you're not using them. And you could consider applying the scope of middleware across your app.

PR #477, opened 2019-01-28
PT MIA 1018 - Christopher Vazquez

cvazquez08:
Functionality compeleted - need to work on styling

PLEASE HELP WITH PLAY BUTTON ON AUDO FILES <33


ta-web-mia:
Nice! And don't worry about the little preview buttons, you're not the only one it affected. You got the code right which is what matters

PR #523, opened 2019-02-13
Monroe Getz FTWD Miami

Mgetz10:


ta-web-mia:
Good job! You're pulling the object values from the API, and using their unique IDs to load webpages for the specific albums. The albums and tracks webpages are looking good. The last step, if you get around to it, is to try and implement the 30 second track demo using an audio tag. Also fix the landing page, it scared the crap out of me 😂 

Keep up the good work 👍

PR #524, opened 2019-02-13
Henry Doce & Guillermo Aiken / FTWD - MIA

aikengx:


ta-web-mia:
So far so good! What you have now only needs some styling, and it should be good. The next step is to route to the albums (and eventually tracks) using object IDs. It will be important that you use the .getArtistAlbums() method that's documented in the spotify-web-api-node documentation: 

https://www.npmjs.com/package/spotify-web-api-node

Once you're able to pull the albums and tracks from the API (which is very similar to what you did for the artists), you'll be all set. Keep up the good work 👍

PR #526, opened 2019-02-15
nunna&matt.FTWD miami

nunna19:


ta-web-mia:
Well done! The main purpose of this assignment was to reference specific objects and values from the API, and to use the Spotify API specific methods, and you did extremely well in that regard. Implementing additional CSS isn't a top priority, so make sure to focus more on the current material. Keep up the good work 👍

PR #529, opened 2019-02-19
FTWD Miami 2019

MikeTroianello:
CSS is not perfect

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,711 @@
+{
+  "name": "lab-express-spotify",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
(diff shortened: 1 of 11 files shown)


ta-web-mia:
Well done! You're pulling from all of the models and using the Spotify API methods properly. As you mentioned, the CSS could definitely be improved upon. In the future, I would recommend using some combination of both bootstrap and Flexbox properties to make a fairly responsive webpage without have to really put that much work into it. Here's a good reference for all things flexbox:

https://css-tricks.com/snippets/css/a-guide-to-flexbox/

You accomplished the main purpose of this assignment extremely well, keep up the good work 👍

PR #552, opened 2019-04-03
[MAD PT] TERRAUN

RedBalloon24:
dificultad: media-alta
comentario: good exercise to practice what we've learned so far


ta-web-mad:
Muy buen trabajo Terraun, todo funciona perfectamente!

#checked #closed #GioƦgio🇮🇹

PR #565, opened 2019-04-10
PAR arnaud alex

alexandfox:
all pages + backend!


starter_code/app.js
```diff
@@ -1,26 +1,85 @@
...
+  res.render("index");
+});
+
+app.get("/artists", (req, res) => {
+  // console.log(req.query.artistName);
+  spotifyApi
+    .searchArtists(req.query.artistName)
+    .then(data => {
+      // console.log("The received data from the API: ", data.body);
+      let artists = data.body.artists.items;
```
ta-web-paris: Here you can create a response.locals instead of a let variable. Response local variables scoped to the request, and therefore available only to the view(s) rendered during that request / response cycle. More efficient

starter_code/app.js
```diff
@@ -1,26 +1,85 @@
...
+    });
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
+  spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .then(data => {
+      // res.json(data);
+      // console.log("The received data from the API: ", data.body);
+      let albums = data.body.items;
```
ta-web-paris: same if you want to = res.locals

starter_code/app.js
```diff
@@ -1,26 +1,85 @@
...
+});
+
+app.get("/tracks/album/:idAlbum", (req, res, next) => {
+  // return console.log(req.params);
+  spotifyApi
+    .getAlbumTracks(req.params.idAlbum)
+    .then(data => {
+      // res.json(data);
+      // console.log("The received data from the API: ", data.body);
+      let tracks = data.body.items;
```
ta-web-paris: same if you want to = res.locals

starter_code/app.js
```diff
@@ -1,26 +1,85 @@
...
+      // console.log(albums);
+      res.render("albums", { albums });
+    })
+    .catch(err => {
+      console.log("The error while searching artists occurred: ", err);
+    });
+});
+
+app.get("/tracks/album/:idAlbum", (req, res, next) => {
+  // return console.log(req.params);
```
ta-web-paris: it is not mandatory at all but here if you want you can destructure you code: 
`const { idAlbum } = req.params `
and then you only call idAlbum as an argument.

ta-web-paris:
GOOD GOOD GOOD

PR #566, opened 2019-04-10
Paris, Miguel Bento && Gerard Joubi

MiguelIronHack:


starter_code/app.js
```diff
@@ -1,26 +1,75 @@
...
-
-
-
-app.listen(3000, () => console.log("My Spotify project running on port 3000 🎧 🥁 🎸 🔊"));
+app.get('/', (req, res) => {
+  res.render('home.hbs');
+});
+
+app.get('/artists', (req, res) => {
+  spotifyApi
```
ta-web-paris: you can see sometimes another way to do the query by deconstructing here:
`const { artistName } = request.query;`
then you can only apply artistName as an argument

ta-web-paris:
GOOD GOOD GOOD GUYS

PR #567, opened 2019-04-10
PARIS - Christofer + Paul Lacour

paullacour:
last iteration almost finished :)

### DIFF
--- a/starter_code/app.js
+++ b/starter_code/app.js
@@ -2,7 +2,7 @@ const express = require('express');
 const hbs = require('hbs');
 
 // require spotify-web-api-node package here:
-
+const SpotifyWebApi = require('spotify-web-api-node');
 
 
 const app = express();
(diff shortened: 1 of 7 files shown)


starter_code/app.js
```diff
@@ -14,13 +14,62 @@ app.use(express.static(__dirname + '/public'));
...
+      // res.json(data.body.items)
+      res.render("albums.hbs", {albums : data.body.items});
+    })
+    .catch(err => {
+      console.log("The error while searching albums occurred: ", err);
+    });
+});
+
+app.get('/tracks/:albumId', (req, res, next) => {
+  spotifyApi.getAlbumTracks(req.params.albumId)
```
ta-web-paris: sometimes you might see this syntax;
`const { albumId } = req.params `
so you will just have to pass albumId as the argument of the method (here getAlbumTracks(albumId))
it is the same

starter_code/app.js
```diff
@@ -14,13 +14,62 @@ app.use(express.static(__dirname + '/public'));
...
+    .then(data => {
+      // res.json(data.body);
+      res.render("tracks.hbs", {tracks : data.body.items});
```
ta-web-paris: As well it is possible that you see this type of code:
` response.locals.tracks = data.body.items `
response.locals is an Express.js way of declaring variable in this scope only, just for the view in question.

ta-web-paris:
It is good BOYZ

PR #568, opened 2019-04-10
PARIS SUSAN&DANI

DaniTOLETUM:
We need to style the albums and tracks pages but all functionality is ok.
Home and artist page are styled.


ta-web-paris:
Some Styling YEAH !
Works fine guys, well Played !

PR #569, opened 2019-04-10
Yang Mimi - Paris 

mimiamsa:


starter_code/app.js
```diff
@@ -1,26 +1,100 @@
...
+      res.render("artists", { artists: data.body.artists.items });
+      // ----> 'HERE WHAT WE WANT TO DO AFTER RECEIVING THE DATA FROM THE API'
+    })
+    .catch(err => {
+      console.log("The error while searching artists occurred: ", err);
+    });
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
+  console.log(req.params);
```
ta-web-paris: another syntax here would have been to deconstruct and only pass artistId as an argument:
`const { artistId } = req.params
.getArtistAlbums(artistId)
`
it is the same, just another way of doing it !

ta-web-paris:
Good Good GUYS 💪

PR #601, opened 2019-04-24
MIA-FTWD-APR-Justin Velez

jvelez523:


MarcosPalaciosIII:
Excellent job on the spotify api. Knowing how to work around the info that you receive from an api will be very helpful because you never know what structure a database may have. Therefore the info can be easy to interpret or complex. Also remember to have your id and secret for any api saved on your .env file so that no one can get ahold of it and do something inappropriate with your account info.

Keep up the excellent work.

PR #602, opened 2019-04-24
MIA FTWD ANGHY POMAR

apomarn:


MarcosPalaciosIII:
Good job on the spotify api. Knowing how to work around the info that you receive from an api will be very helpful because you never know what structure a database may have. Therefore the info can be easy to interpret or complex. The track previews did not work showing an unknown source, make sure to log everything to make sure you are getting the right info prior to using it.

Keep up the good work.

PR #605, opened 2019-04-24
MIA FTWD APR - Katy Brown

katybrown91:


MarcosPalaciosIII:
Good job on the spotify api. Knowing how to work around the info that you receive from an api will be very helpful because you never know what structure a database may have. Therefore the info can be easy to interpret or complex. Also remember to have your id and secret for any api saved on your .env file for privacy. Also you have to remember to set you .env file as an ignored file or others can still see what you have in that file.

Keep up the good work.

PR #654, opened 2019-07-05
MIA-FTWD-Sergio and Eduardo

sergiovelo17:


ta-web-mia:
Great job! Although, for the search random artist, I couldn't get the songs to actually play. It works if you search for a specific artist, and that's what counts.

PR #743, opened 2019-09-11
MIA-FTWD-Elba Chimilio

elba3184:


ta-web-mia:
Well, you got the tracks to play, so good job! There is a glitch I encountered, where upon the first time I clicked an album, all the tracks played at once. I've never gotten to hear a whole album at the same time, so maybe that's a feature?

PR #796, opened 2019-11-08
[BER-WDFT-William & Michael]

whbrown:


starter_code/.env
```diff
@@ -1,5 +1,5 @@
 # the Spotify api key and secret should be saved in this file, 
 # so when ready uncomment following lines and add them
 
-# CLIENT_ID= your clientId goes here
-# CLIENT_SECRET= your clientSecret goes here
\ No newline at end of file
+CLIENT_ID= f12e334c34f04cd19e64368ba8b83212
```
itstheandre: You guys still haven't learned this yet, but this should never be public. whenever you have time, go to spotify and reset these hashes. and then add a .gitignore file to your root folder with .env in it.

This way this sensitive data will never be passed

PR #807, opened 2019-11-10
[BER-WDFT-102019-FABIAN]

fabianschu:


starter_code/app.js
```diff
@@ -15,14 +15,79 @@ app.use(express.static(__dirname + '/public'));
...
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
+  
+  console.log(req.params.artistId);
+  spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .then(data => {
+      console.log(data.body);
+      let albumsResponse = data.body;
```
DalinaLeila: store this in one variable only and then pass it on. No need to create two variables :)

starter_code/app.js
```diff
@@ -15,14 +15,79 @@ app.use(express.static(__dirname + '/public'));
...
+
+app.get("/tracks/:albumId", (req, res, next) => {
+  
+  console.log(req.params);
+  
+  spotifyApi
+    .getAlbumTracks(req.params.albumId)
+    .then(data => {
+      console.log(data.body);
+      let tracksResponse = data.body;
```
DalinaLeila: see above

PR #892, opened 2020-01-31
[BER-FT-012020] Jacqueline C.

chenjacqueline:
Done + styling!

### DIFF
--- a/starter-code/public/styles/style.css
+++ b/starter-code/public/styles/style.css
@@ -1,0 +1,128 @@
+.background {
+  height: 100vh;
+  width: 100vw;
+  background-image: url("/images/spotify-background.jpeg");
+  background-size: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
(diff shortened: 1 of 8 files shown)


starter-code/public/styles/style.css
```diff
@@ -0,0 +1,128 @@
+.background {
+  height: 100vh;
+  width: 100vw;
+  background-image: url("/images/spotify-background.jpeg");
+  background-size: 100%;
```
fenrew: This may look a bit better if you're making the webpage responsive? Background-size: 100% 100%; If you use  Background-size: 100%, then it should be considered using background-repeat: no-repeat; and height: auto;
Background-size: cover; also looks good to me.
You'll see the difference on smaller screens.

chenjacqueline:
Ah! This is a good feedback, I was debating btw 100% or cover for a while.
Thank you :)

fenrew:
Always! Keep in mind that there is a difference between background-size: 100% 100%, and background-size: 100%, :)

PR #895, opened 2020-01-31
[BER-FT-012020] Mia L.

mialav:


starter-code/public/styles/style.css
```diff
@@ -0,0 +1,104 @@
...
+  margin: 0;
+  padding: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+/* FRONT PAGE */
+
+.front-page {
+  background-image: url(/images/spotify-background.jpeg);
+  background-size: 100%;
```
fenrew: This may make your webpage look a bit wierd if the screen of the user is a bit smaller than yours, so consider using background-size: 100% 100%, or background-size: cover.

But in the end, you can choose yourself how to style your webpage!:)

PR #897, opened 2020-01-31
[BER FT 01-2020] Konstantin

kdauer:


starter-code/public/styles/style.css
```diff
@@ -0,0 +1,128 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+
+  background-color: rgba(255, 255, 255, 0.5);
+}
+
+form input {
+  min-width: 500px;
```
fenrew: Nice! Responsive all the way is really good practice:)

PR #898, opened 2020-01-31
[BER-FT] Elsa

ElsaFil:
- I am not sure if I am using the correct dynamic links (should it be routes or queries?)
- I created a link under each picture using an <a> tag with padding. And then I put a <div> around it for the background gray colour. But I never managed to position them correctly. Any idea how I could do it? Or does it need a whole different approach? 🤷‍♀️

### DIFF
--- a/starter-code/app.js
+++ b/starter-code/app.js
@@ -1,18 +1,70 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
+const SpotifyWebApi = require("spotify-web-api-node");
 
 const app = express();
 
-app.set('view engine', 'hbs');
-app.set('views', __dirname + '/views');
-app.use(express.static(__dirname + '/public'));
+app.set("view engine", "hbs");
+app.set("views", __dirname + "/views");
+app.use(express.static(__dirname + "/public"));
+hbs.registerPartials(__dirname + "/views/partials");
 
 // setting the spotify-api goes here:
+const spotifyApi = new SpotifyWebApi({
+  clientId: process.env.CLIENT_ID,
+  clientSecret: process.env.CLIENT_SECRET
+});
+
+// Retrieve an access token
+spotifyApi
+  .clientCredentialsGrant()
+  .then(data => spotifyApi.setAccessToken(data.body["access_token"]))
+  .catch(error =>
+    console.log("Something went wrong when retrieving an access token", error)
+  );
 
 // Our routes go here:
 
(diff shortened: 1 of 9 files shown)


starter-code/public/styles/style.css
```diff
@@ -0,0 +1,112 @@
...
+  padding: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.main {
+  background-image: url("/images/spotify-background.jpeg");
+  background-size: cover;
+  height: 100vh;
+  display: flex;
+  flex-direction: column;
```
fenrew: This line doesn't really do anything in your code;)

starter-code/public/styles/style.css
```diff
@@ -0,0 +1,112 @@
...
+  align-items: center;
+  justify-content: center;
+}
+form {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  width: 700px;
```
fenrew: I'd personally assign the width in vw; eg: 50vw, and then set the min-width to 400px or something

starter-code/public/styles/style.css
```diff
@@ -0,0 +1,112 @@
...
+  height: 200px;
```
fenrew: Height as well i guess: height: 10vw (or whatever), and min-height: 100px

starter-code/app.js
```diff
@@ -1,18 +1,70 @@
...
+    .then(data => {
+      res.render("artist-search-results.hbs", {
+        artists: data.body.artists.items
+      });
+    })
+    .catch(err =>
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+app.get("/albums", (req, res, next) => {
```
fenrew: Here it is better to use: app.get("/albums/:id", .....

and then use req.params.id    this will give you back the id and it looks better. req.query is meant for the search function, while req.params are neater for getting a particular item that is clicked.

starter-code/app.js
```diff
@@ -1,18 +1,70 @@
...
+  spotifyApi
+    .getArtistAlbums(req.query.artist)
+    .then(data => {
+      res.render("albums.hbs", { albums: data.body.items });
+    })
+    .catch(err => {
+      console.log("The error while searching albums occurred: ", err);
+    });
+});
+app.get("/tracks", (req, res, next) => {
```
fenrew: Same here I assume

starter-code/views/artist-search-results.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="artists">
+    {{#each artists}}
+    <div class="box">
+        {{> tile this}}
+        <div class="gray-bg">
+            <a href="/albums?artist={{id}}">View Albums</a>
```
fenrew: I believe it's better to do it like this: <a href = "/arbums/{{id}}

PR #899, opened 2020-01-31
[BER - FT - WD - 0120] Jonathan 

JonathanSaudhof:


fenrew:
For future reference: Make sure that the .gitignore is put into the same folder or any consequent subfolders as your node_module. This will prevent your node modules to be uploaded into github. Currently it says that you added 200 000 new lines of code to the Lab. Please ask me if this is unclear:)

Your spotifyApi clientId and clientSecret is currently accessible to everyone in the whole world (just so you know). The right way to add secret key and id's is to make a file called: .env
In this file you can add a plain text something along the lines of:
CLIENT_ID=096b5cajoieajfae9u98faepk90ilea
CLIENT_SECRET=9joeja98evaj0ejapjeov9eua00u

and then in your code you can write:
const spotifyApi = new SpotifyWebApi({
  clientId: process.env.CLIENT_ID,
  clientSecret: process.env.CLIENT_SECRET
});

and then in your git ignore you can tell it to ignore the .env file. What will happen then is that git will not upload your client Id or client secret so no one can enter and screw up your account. Once again, tell me if this is unclear!

The artist-search-results.hbs may look a bit messy, and in a lot of cases it can be better to divide it into several different files, but then again: It's all about preference and your way may be better for you in the end.

Overall: I think this is really good!! You're using some lines in the backend which I am unfamiliar with, like: app.engine. But it seems like you very much know what you're doing, so you can for sure  experiment and have some fun with the code!:)

JonathanSaudhof:
Hi Markus!
I prefer express-handlebars over handlebars. You can set a directory for
layouts! That's basically it :D.
It seems to be a bit easier for me. But overall its almost the same.

Thx for your feedback. :)
Jonathan

Am Mo., 3. Feb. 2020 um 12:16 Uhr schrieb Markus Heldrup <
notifications@github.com>:

fenrew:
Aha, that's awesome!! I like how interested you are in programming and how you keep finding new ways or frameworks to play around with, keep that up! :D

PR #902, opened 2020-01-31
[BER FT]Tim 

timbotimber:
Done the functionality, not the styling though. Feel like I got the concept by the end of the lab but it was really tricky. Found it quite frustrating as I felt like I had the concepts in the class then found it tricky to implement. Hope some reading over the weekend will help!

### DIFF
--- a/starter-code/app.js
+++ b/starter-code/app.js
@@ -1,18 +1,82 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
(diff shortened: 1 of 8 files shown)


DalinaLeila:
I know how you are feeling, totally normal! Don't forget that you only learned the concepts a few hours before you did the lab. You will have to do this now for the next days for every lab and also for your project so the concepts and the syntax will become easier with every time you code. 👍 

If you need a recap or help, let me know we can go over the material in a break :)

PR #904, opened 2020-01-31
[BER-FT] - Marcel 

Marsu-R:


starter-code/views/tracks.hbs
```diff
@@ -0,0 +1,19 @@
+<section class="tracks-page">
```
DalinaLeila: you could also do an HTML table here :)

starter-code/views/artist-search-results.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="results-container">
+    {{#each artistList}}
+    <div class="row">
+        <div>
+            <img class="image" src=" {{{ this.images.0.url }}} " alt="image of {{{ this.name }}}">
```
DalinaLeila: use only 2 `{{}}` not 3 `{{{}}}` here and for all hbs files. The only time when you want to use three is if you want to include html tags (E.g for the body in the layout hbs)

PR #906, opened 2020-02-02
[BER-FT-012020]) - Saeka

saekaisozumi:


starter-code/views/tracks.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="header">
```
DalinaLeila: you could use the HTML table here

PR #909, opened 2020-02-03
[BER - FT] - Nora Nicklis

Norizinus:


Norizinus:
btw, I didn't get the dotnev thing to work, so for now I put the client ID and secret into a variable. That worked. Whenever I tried using the other way, it didn't any more. Maybe you can have a look at that :)

PR #914, opened 2020-02-03
#wd-ft-ber-01-20 martin done

martinschmal:
but with very basic design


martinschmal:
> I am not able to run your lab. I am having problems with the dependencies.
> Was it running on your side correctly?

yes. Double checked and pushed again.

PR #968, opened 2020-03-08
done and now working on cssand styles

teo-23:
I might need a recap on CSS because it takes me forever for basic steps :)

### DIFF
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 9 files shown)


views/index.hbs
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Homepage</title>
+  <link rel="stylesheet" href="/styles/styleHome.css">
+</head>
+<body>
+  <div class="search">
```
LloydC: Your form element  doesn't work because the label, input and button need to be nested within it so that you can submit the artist name properly. You should remove the unnecessary div surrounding those elements as well.

PR #974, opened 2020-03-12
Done! [AMS-PT-012020] - Alex

alexpanko:


views/index.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="bgImage">
+    <div class="searchBox">
+        <form action='/artist-search' method='GET'>
+            <input class="searchInput" type="text" name="search">
```
LloydC: Just wanted to mention that there's a 'required' attribute which you can use on your input elements. It's useful for data validation before the form gets submitted on the back end. Nice work!

PR #1030, opened 2020-04-19
Berlin WD Dioni👨‍💻

dionifss:
All done


dionifss:
with a image for the broken images, so it doesnt break the design ;)

dionifss:
with  artist name in the lists ;)

PR #1101, opened 2020-06-26
BER WDFT - Leo Costa

Leoprogramming:
Main iterations done - working. ✓

No styling and probably parts are missing.

Found this lab tough, but very rewarding. Was able to (mostly) do it with help from colleagues/teacher, very unlikely I would have finished it on my own, but I feel like I'm learning a lot.

Will see if I manage to go over it on the weekend again and maybe do the styling.

### DIFF
--- a/views/tracks.hbs
+++ b/views/tracks.hbs
@@ -1,0 +1,5 @@
+{{#each track}}
+  <p>{{name}}</p>
+  <img src="{{this.images.1.url}}">
+  <a href="/album/{{this.id}}"></a>
+{{/each}} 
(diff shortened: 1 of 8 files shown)


views/tracks.hbs
```diff
@@ -0,0 +1,7 @@
+{{#each track}}
+  
+  <p>{{name}}</p>
+  <img src="{{this.images.1.url}}">
+  <a href="/album/{{this.id}}"></a>
+
```
Jeff-Moraes, line 6: só faltou add a tag de `audio` desta forma para mostrar o player
```<audio controls src="{{preview_url}}" />```

PR #1149, opened 2020-08-01
[REMOTE-PT-06-20] Raquel Nascimento & Debora Crescenzo 

NubivagantYansa:
Done CSS INCLUDED! Great team 👯 

Questions ✋ :
1. How do we set a default image when we don't receive one from the API?
2. Raquel and I have the same code but for her the styles for the albums and tracks is not showing - how's that even possible?!?! CSS **dark** magic 🙃  - SORTED
3. How do we make the name of the artists appear on the album heading e.g. 'Album for: Beatles' (see album.hbs) 

Thaaaaaaaaanks

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,18 +1,71 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
+const SpotifyWebApi = require("spotify-web-api-node");
 
 const app = express();
 
-app.set('view engine', 'hbs');
-app.set('views', __dirname + '/views');
-app.use(express.static(__dirname + '/public'));
+app.set("view engine", "hbs");
+app.set("views", __dirname + "/views");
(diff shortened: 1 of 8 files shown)


itstheandre:
Hey Debora , please send this also to Raquel 
Regarding a default image, you would have to do that with Javascript. You would have to check if the image is broken somehow, but that probably wouldn't be done using this kind of project. You could add an "event listener" on the image tag and check if there is an error - there is an onError event listener on the image tag" and dynamically change the image.

Regarding the css, I would have to check the code, but the most probable causes are: browser (it can be a browser compatibility). The other reason is probably a typo somewhere. A missing slash, or a misnamed folder. I would have to look at it and check it myself.

Regarding the name, I would first do a call to spotifyApi to get the artist by the params. There you would be able to get the artists name.
Store the name in a variable.
Something like this:
```js
spotify.getArtist(req.params.id).then(artistData => {
    const artist = artistData.body.name;
    spotify.getArtistAlbums(req.params.id).then(data => {
        res.render("albums", {album: data.body.items, artist: artist });
```
Here you would be sending the artist also to the frontend

PR #1162, opened 2020-08-03
[REMOTE WDPT-0602]-Nuno Pacheco

nuno-pacheco:
It was great to have time during class to do the lab.
My pair was Willem, we worked separately but at same time, interacting when needed.


nuno-pacheco:
It was great to have time during class to do the lab.
My pair was Willem,  we worked separately but at same time, interacting when needed.

PR #1165, opened 2020-08-04
[WDPT REMOTE 0602] Lynn W-B

lynn-w-b:
I didn't spend a long time on the styling because the CSS was driving me crazy!!

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,18 +1,71 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
-
-// require spotify-web-api-node package here:
+const express = require("express");
+const hbs = require("hbs");
+const SpotifyWebApi = require("spotify-web-api-node");
 
 const app = express();
 
-app.set('view engine', 'hbs');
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -1,18 +1,71 @@
...
 
 // Our routes go here:
+app.get("/", (req, res) => {
+  res.render("index");
+});
+app.get("/artist-search", (req, res) => {
+  spotifyApi
+    .searchArtists(req.query.artistName)
+    .then(data => {
+      const { name, images, id } = data.body.artists.items[0];
```
valmgisbert, line 34: Nice deconstruction! The only thing here is that we not only want it to show the specific artist we're looking for, but all that have a similarity with the name if we don't write it as exactly as it is; by determining the first item, you're limiting the result to just one. This would also mean that you should adapt your artists file to multiple results and not just one.

valmgisbert: Another recommendation is that since handlebars in itself helps us control what information we want to extract from a big block of data, to aid yourself with the helpers you can try grabbing the path that you have to get your response inside a variable which then you turn into an object to use in your handlebars files.

PR #1166, opened 2020-08-04
[REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:
I had problems with handlebars

### DIFF
--- a/views/albums.hbs
+++ b/views/albums.hbs
@@ -1,0 +1,5 @@
+<div>
+  {{#each albums}}
+    {{> album}}
+  {{/each}}
+</div>
--- a/views/partials/artist.hbs
+++ b/views/partials/artist.hbs
@@ -1,0 +1,3 @@
+<a href="/albums/{{id}}}">View Albums</a>
+<p>{{name}}</p>
+<img src="{{images.url}}" alt="">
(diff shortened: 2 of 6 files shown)


views/partials/artist.hbs
```diff
@@ -0,0 +1,3 @@
+<a href="/albums/{{id}}}">View Albums</a>
+<p>{{name}}</p>
+<img src="{{images.url}}" alt="">
```
valmgisbert, line 3: Remember that you might get multiple images with your request, you should pinpoint the first image of the array.

views/albums.hbs
```diff
@@ -0,0 +1,5 @@
+<div>
+  {{#each albums}}
+    {{> album}}
```
valmgisbert, line 3: I see you calling this file as your partial, but you don't have anything in that file; since it doesn't have a template to use, your partial won't work.

valmgisbert:
I would tell you to use a single partial for both pages with just name and first image, since with the button/anchor you need to change the links directing to the next page, even if both look for id. My first recommendation is to write it without the partials, and if it works correctly, refactor with your partials. Also, make sure to clean your console.logs(). Other than that, good job!

PR #1175, opened 2020-09-03
[PTWD-AMS062020] Mahmut Sarcan

mhmtdl:


app.js
```diff
@@ -12,7 +13,61 @@ app.set('views', __dirname + '/views');
...
+    // ----> 'HERE WHAT WE WANT TO DO AFTER RECEIVING THE DATA FROM THE API'
+    res.render("artist-search-results",{data:data.body.artists.items})
+  })
+  .catch(err => console.log('The error while searching artists occurred: ', err));
+})
+
+
+app.get('/albums/:artistId', (req, res, next) => {
+  // .getArtistAlbums() code goes here
+  const {artistId:artistId} = req.params;
```
guido-an, line 49: This is enough: `const { artistId } = req.params;`

guido-an: Same in the next route

PR #1177, opened 2020-09-04
[WDFT Berlin 08.20] Hamlet Hayrapetyan

Prime-arm:


views/albums.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="albums-container">
+    {{#each }}
+    <img src="{{}}" alt="">
```
timbotimber, line 3: Here, you could use the same structure as the image link in the artist-search-results.hbs file.

timbotimber (commented):
Looks good! Think you guys get the concepts pretty well

PR #1178, opened 2020-09-04
WEBDEV-BERLIN-KevinKlein (done without css)

pkevinklein:


views/albums.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="album-container">
+{{#each searchedArtist}}
+<h2>{{name}}</h2>
+{{/each}}
+
```
Tyouxik, line 5: You are trying to use searchedArtist, but you are not passing this information in your route.

Tyouxik (commented):
Nice job. Continue like this !

views/albums.hbs
pkevinklein: Thanks a lot for your feedback! The artist's name has been implemented 🔢  in the albums page

PR #1179, opened 2020-09-04
WDFT Berlin Mira

vomira:


app.js
```diff
@@ -12,7 +13,57 @@ app.set('views', __dirname + '/views');
...
+});
+
+app.get('/albums/:artistId', (req, res) => {
+    spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .then(albums => {
+        res.render('albums', {
+            albums: albums.body.items
+        })
+    })
```
Jeff-Moraes, line 53: It is a good approach to always add the ``.catch`` when dealing with promises ;)
```
.catch((err) =>
      console.log("The error while searching artists occurred: ", err)
    );
```

Jeff-Moraes (commented):
Super well done!!
Great that you solved all the interactions and also added the style, it looks even better than the example. ;)
I've added just one comment.
Nice work!! Congrats Mira!

PR #1182, opened 2020-09-05
WDFT-Berlin-Tolga Kara

TolgaKara:
The Design is not great but my goal were the functionality.

- I need to check async and await because in django I never used it and it doesn't felt that I need it.

### DIFF
--- a/app.js
+++ b/app.js
@@ -10,6 +10,7 @@ const app = express()
 
 app.set("view engine", "hbs")
 app.set("views", __dirname + "/views")
+hbs.registerPartials("/views/partials")
 app.use(express.static(__dirname + "/public"))
 
 // setting the spotify-api goes here:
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -1,18 +1,73 @@
...
+	console.log()
+	await spotifyApi
+		.getArtistAlbums(req.params.id)
+		.then((val) => {
+			console.log(val.body.items)
+			res.render("albums", { albums: val.body.items })
+		})
+		.catch((err) => console.log("Something went wrong by getting the Albums: ", err))
+})
+
```
Tyouxik, line 61: I love that you try to implement Async/Await function.
It is an alternative to .then, and .catch, in your code you combine both.
You could have done this for example:
app.get("/artist-search", async (req, res) => {
  try {
    let allArtists = await spotifyApi.searchArtists(req.query.name);
    console.log(allArtists);
    await res.render("artist-search-results", {
      artists: allArtists.body.artists.items,
    });
  } catch (e) {
    console.error(e);
  }
});

Tyouxik (commented):
Great job !

app.js
TolgaKara: Ohh I thought the .then and .catch is similar to try{} catch{}

PR #1190, opened 2020-09-07
Webdev aug2020 Esther Wang 

sewangco:


views/artist-search-results.hbs
```diff
@@ -0,0 +1,13 @@
...
+<p>
+    {{!-- {{#each this.images}}
+{{url}}
+    {{/each this.images}} --}}
+</p>
+<p>
+    {{#each this.name}}
+    {{/each}}
+</p>
+
```
Tyouxik, line 12: You don't need to iterate through this.name, each artist has one name.

you only need {{name}}

Tyouxik (commented):
Good job!

PR #1192, opened 2020-09-07
FT - Berlin - Veronica De Leon

veronicadeleonh:


app.js
```diff
@@ -1,18 +1,93 @@
...
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
+  // .getArtistAlbums() code goes here
+  // console.log("id", req.params.artistId);
+  spotifyApi
+    .getArtistAlbums(req.params.artistId, {
+      limit: 5,
+      offset: 1,
+    })
```
Jeff-Moraes, line 43: Great that you also used some query to limit the data ;)

app.js
```diff
@@ -1,18 +1,93 @@
...
+    .then(
+      function (data) {
+        console.log("This is the album object", data.body.items);
+        res.render("tracks", { tracksArr: data.body.items });
+      },
+      function (err) {
+        console.log("Something went wrong!", err);
+      }
+    );
```
Jeff-Moraes, line 73: In this case, the catch method is a better approach to handle the error, instead of using a callback function.
```
.catch((err) =>
   console.log("Something went wrong!", err)
);
```

Jeff-Moraes (commented):
Super well done Veronica!!
Great that you solved all the interactions and also added some style.
I've added just a few comments.
Nice work!! Congrats!

PR #1193, opened 2020-09-07
Completed lab

nadjacreischel:


app.js
```diff
@@ -13,6 +15,67 @@ app.use(express.static(__dirname + '/public'));
...
+    res.render('home')
+});
+
+app.get('/artist-search', (req, res) => {
+    // console.log(data.body.artists.item)
+    spotifyApi
+    .searchArtists(req.query.title)
+    .then(data => {
+    console.log('Search artists', data.body.artists.items);
+    res.render('artistSearch', data.body)
```
Jeff-Moraes, line 50: Here instead of just send the data on this way, it is a better approach to send as an object and give to the data a proprietary name
```res.render("albums", { artists: data.body.items });```

because, imagine that you want to send more than one data, so it will be better that each data has a name.

File: app.js [hunk 2]
```diff
@@ -13,6 +15,67 @@ app.use(express.static(__dirname + '/public'));
...
+  })
+})
+  
+app.get('/albums/:artistId', (req, res, next) => {
+    console.log(req.params.artistId)
+    spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .then(data => {
+          console.log('Artist albums', data.body);
+          res.render('albums', data.body)
```
Jeff-Moraes, line 60: same here ;)
```res.render("albums", { albums: data.body.items });```

File: views/albums.hbs [hunk 3]
```diff
@@ -0,0 +1,5 @@
+{{#each items}}
```
Jeff-Moraes, line 1: then her instead of items it will be albums.

Jeff-Moraes (commented):
Nicely done Nadja!!
Great that you solved all the interactions.
I've just added a few comments about the data that you sending to the view.
Nice work so far, keep it up! ;)

PR #1194, opened 2020-09-07
WDFT BER 082020 - Gilad Tsabar

giladt:


timbotimber (commented):
Real nice man. Looks like you really get it. If you want an extra challenge you could try using async/await functions for your promises in the routes file. It's a really nice upgrade and makes the code a little easier to read

PR #1251, opened 2020-11-06
[FTAMS102020] Rich Alossi

alossix:


LloydC:
Good job Rich!

I have nothing to add, your solution doesn't have anything missing or extra.

PR #1252, opened 2020-11-06
[FTAMS102020] Nina van den Braak

ninsajala:
Most important part is finished


LloydC:
Good work Nina!

The only advice I would give you is to destructure the data you're receiving from Spotify on the Back End instead of the Front End.  A general rule is that you only need to send the data that the user requires on the page, sending data that doesn't serve any purpose is unnecessary.

So instead of sending the entire `data` object from your routes to the front end, you should create another variable that would store only the data you need to send and then send that.

i.e:
```
app.get(`/artist-search`, (req, res, next) => {
    spotifyApi
        .searchArtists((req.query).artist)
        .then(data => {
         const artistsList = data.body.artists.items;
            res.render(`artist-search-results`, {
                artistsList
            })
        })
        .catch(err => console.log('The error while searching artists occurred: ', err));
});

PR #1254, opened 2020-11-06
[FTWEBAMS102020]Mahalakshmi

mahavinay:
going well , on iteration 4

### DIFF
--- a/views/albums.hbs
+++ b/views/albums.hbs
@@ -1,0 +1,10 @@
+<ul>
+    {{#each albums}}
+    <li> {{this.name}}</li>
+       <br> 
+       <img src="{{this.images.2.url}}" alt="albumcover">
+       
+       <a href="/tracks/{{this.id}}">View tracks</a>
+    {{/each}}
+    <br>
+</ul>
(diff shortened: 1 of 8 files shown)


views/albums.hbs
```diff
@@ -0,0 +1,10 @@
+<ul>
```
LloydC, line 1: Everything works fine, the only issue here is how you're nesting other elements than `<li>` inside the `<ul>`.

You should only have `<li>` items, you can nest other elements within it like this for instance
```
<ul>
    {{#each albums}}
    <li> 
     <div>
       <span>{{this.name}}</span>
       <br> 
       <img src="{{this.images.2.url}}" alt="albumcover">     
       <a href="/tracks/{{this.id}}">View tracks</a>
       {{/each}}
       <br>
     </div>
  </li>
</ul>

File: views/artist-search-results.hbs [hunk 2]
```diff
@@ -0,0 +1,15 @@
+<div class="artists">
```
LloydC, line 1: You should have a `<ul>`element here as well

views/tracks.hbs
```diff
@@ -0,0 +1,14 @@
+<ul>
+    <h1>Tracks</h1>
```
LloydC, line 2: Your heading should be before the `<ul>` instead of inside.

PR #1255, opened 2020-11-06
[FTAMS102020] Emma Northridge

emmanorthridge:
Functionality is working as expected.
All iterations completed


LloydC:
Good work Emma!

I don't see anything off with your solution, just want to mention that you don't need the '.hbs' extension as long as you have set handlebars in your express settings `app.set("view engine", "hbs");`

PR #1257, opened 2020-11-07
[FTAMS102020] Alex Kantsadis

teof95:
done without styling

### DIFF
--- a/views/albums.hbs
+++ b/views/albums.hbs
@@ -3,7 +3,8 @@
     {{#each albums}}
 <li> {{this.name}}</li>
 <br>
-<img src={{this.image.1.url}} alt="album-cover">
-<a href="">View Tracks</a>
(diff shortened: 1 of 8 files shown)


views/albums.hbs
```diff
@@ -0,0 +1,10 @@
+<hi>In albums</hi>
+<ul>
```
LloydC, line 2: Everything works fine, the only issue here is how you're nesting other elements than <li> inside the <ul>.

You should only have <li> items, you can nest other elements within it like this for instance
```
<ul>
    {{#each albums}}
    <li> 
     <div>
       <span>{{this.name}}</span>
       <br> 
       <img src="{{this.images.1.url}}" alt="{{this.name}}">     
       <a href="/tracks/{{this.id}}"><button type="button"View tracks</button></a>
       {{/each}}
       <br>
     </div>
  </li>
</ul>

PR #1258, opened 2020-11-07
[FTAMS102020] Irene Asensio

ihatepineapple:


views/tracks.hbs
```diff
@@ -0,0 +1,15 @@
+<h1>tracks</h1>
+
+{{#each tracks}}
```
LloydC, line 3: I think it would be better to nest `{{#each tracks}}` within the `<ul>`.

The way it's setup now makes it generate `<div class="tracklist-container">` multiple times instead of making multiple `<li>` elements.

ihatepineapple: Hi Lloyd, yes that was my precise idea, for styling reasons

PR #1302, opened 2020-11-13
[WDFTRMT- 12112020] - Milena Martinez 

MilenaMar:
Lab completed.
took me a while figured how was structured the API from Spotify (10000 console.log), 
after that everything was more easy

### DIFF
--- a/views/artist-search-results.hbs
+++ b/views/artist-search-results.hbs
@@ -1,0 +1,11 @@
+<div class="search">
+    {{#each artist}}
+    <div class="card">
+        <img class="card-img-top" src={{images.[0].url}} alt="artist album">
+        <div class="card-body">
+            <h4 class="card-title">{{name}}</h4>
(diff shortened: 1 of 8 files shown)


filipe-freire (commented):
Awesome job Milena! Working with API's can always be a bit tricky, but this here is a job well done, congrats! (and nice work with the styling!) 🥇

views/artist-search-results.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="search">
+    {{#each artist}}
+    <div class="card">
+        <img class="card-img-top" src={{images.[0].url}} alt="artist album">
```
filipe-freire, line 4: When array items are objects, you can just access that object using the dot notation in Handlebars. Ex: images.0.url

MilenaMar: Thanks for the note Filipe 😃will put it in practice from now on

PR #1309, opened 2020-11-13
WDFTRM 131120 - Francisco Melo

xikz:
Finished.
Issue with styling CSS:: Don't understand why by using flex im getting a huge horizontal scroll. Why are the all the divs in one single row? (e.g. in the artist-search-results (if there are alot of artists)). Is it because of hbs each loop? is there a way araoud this?

### DIFF
--- a/public/styles/style.css
+++ b/public/styles/style.css
@@ -1,0 +1,44 @@
+.tracks-table {
+  margin: 20px;
+}
+
+.tracks-header {
+  display: flex;
+  justify-content: space-between;
+  background-color: black;
+  color: white;
+  padding: 0px 20px;
+}
+
+.track {
+  display: flex;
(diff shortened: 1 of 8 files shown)


public/styles/style.css
```diff
@@ -0,0 +1,44 @@
...
+}
+
+h4 {
+  margin: 10px;
+  width: 400px;
+}
+
+.albums {
+  width: 100vw;
+  display: flex;
```
IH-WebDev-TA-Remote, line 32: Try adding the property: `flex-wrap: wrap;` to this albums class! 👍

IH-WebDev-TA-Remote (commented):
Awesome job Francisco! Very nicely done 🥇 Added a comment regarding your horizontal scrolling problem!

PR #1320, opened 2020-11-14
Completed LAB for Express Spotify API

MADFutcher:
The Tracks do not have a preview_url any more. The API just fills it with null.

I took a screen shot of the playing icon in the example and used this in the listen section


MADFutcher:
Updated it with the Preview URL as i have been shown that some do have the Preview URL :)

PR #1324, opened 2020-11-15
[FTWDRM 2020/11/15] Hugo

h-u-g-o-1989:


filipe-freire (commented):
Nice work Hugo! 🏆 Working with API's can sometimes be tricky, but you did very good! Contrats!

PR #1435, opened 2021-03-14
[WDPT JAN21] Aleksandra Porada: Essential part done (all iterations fixed but no styling)

hola-ola:
I would be interested to learn how to display the name of the artist on the albums page. In other words, how to reach and display a piece of data from the "collection" that was requested for another page.


IH-WebDev-TA-Remote:
Hey Aleksandra! First of all, nicely done, congratulations! 😄 Second, one way to achieve that is to pass that information along with the view and the albums. E.g `res.render("albums", { albums: arrayOfAlbums, artistName: theArtistsNameHere });`. We would have to make another API call in case the artists name was not provided in albums!

PR #1436, opened 2021-03-14
John Rosario [WD RMT PT JAN21] BERLIN

knottykid:
did only part of the styling, but is better than nothing. it was a fun lab!


IH-WebDev-TA-Remote (commented):
All good John! Congratulations on another completed lab, everyday getting better! 😎

PR #1457, opened 2021-04-16
BER-WDFT-MAR2021-Alkin

SirDev97:


wax-bear (commented):
Good stuff Alkin!

The application enables searching for artists, clicking on their albums and listening to their tracks.
There is no CSS implemented yet. This would be helpful to make the page look more user friendly!

Best,

Thomas

PR #1465, opened 2021-04-16
BER - FTMA2021 - Jan Colomer

Jancololo:


app.js
```diff
@@ -12,7 +13,57 @@ app.set('views', __dirname + '/views');
...
+
+
+//Iteration 4: view albums
+app.get('/albums/:artistId', (req, res, next) => {
+    // .getArtistAlbums() code goes here
+    spotifyApi.getArtistAlbums(req.params.artistId)
+        .then(function (data) {
+            console.log('Artist albums', data.body.items[0].artists[0].name);
+            res.render('albums', { artistAlbums: data.body.items })
+        }, function (err) {
```
annawaszak, line 51: I would suggest to make it a "catch" function instead of anonymous one.

annawaszak (commented):
Hi Jan!

Outstanding work on this one, you clearly have a good understanding of the material covered in module 2. Looking forward to seeing you project :)

Thanks,
Anna

app.js
Jancololo: Upss! Sure I'll change that. Thanks for the feedback!

PR #1466, opened 2021-04-16
WDFT BER MAR21 - R.Stein

Chocogrenouille:


wax-bear (commented):
Great stuff Roxi!

The interaction with the API works as expected.
Perhaps you could think of providing a default image in "artist-search" when none is available. Moreover, try to be aware of the consistent use of semicolons and empty lines.

Best,

Thomas

PR #1467, opened 2021-04-16
[FT-WD-BER] -  David Bueno

davidbuenox8:


app.js
```diff
@@ -1,18 +1,68 @@
...
+            res.render('tracks', { tracklists: data.body.items })
+
+        }, function (err) {
+            console.log('Something went wrong!', err);
+        });
+})
 
 // Our routes go here:
+app.get('/', (req, res) => {
+    res.render('layout')
```
ahmedhany82, line 65: Layout is the default page and what you pass to `res.render` is rendered inside the body tag in `layout.hbs`. You do need to render layout separately and this is rendering here is what makes the search bar appears twice in the main page. I would render here a new hbs file (`home.hbs`) and put inside the search bar and background image.

views/albums.hbs
```diff
@@ -0,0 +1,5 @@
+<h1>Albums</h1>
+
+{{#each albumartist}}
+{{> albumCard this}}
```
ahmedhany82, line 4: It is OK to use Partials here but usually they come in handy when the same code is reused in different hbs files.

views/artist-search-results.hbs
```diff
@@ -0,0 +1,5 @@
+<h1>Artists</h1>
+
+{{#each artists}}
+{{> artistCard this}}
```
ahmedhany82, line 4: Same here about the partial.

ahmedhany82 (commented):
Hey David,
Good job for the lab! Please find my review remarks. Feel free to reach out if you have any questions.
All the best!
Ahmed

PR #1473, opened 2021-04-17
[Berlin-WDFT-MAR2021] Marcia Faria Netter

mfndnc:


app.js
```diff
@@ -4,15 +4,108 @@ const express = require('express');
...
+      }
+    })
+    .catch((err) =>
+      console.log('The error while retrieving tracks occurred: ', err)
+    );
+});
 
-app.listen(3000, () => console.log('My Spotify project running on port 3000 🎧 🥁 🎸 🔊'));
+app.listen(port, () =>
+  console.log('My Spotify project running on port 3000 🎧 🥁 🎸 🔊')
```
ahmedhany82, line 110: I think you could mention port 3010 here.

app.js
```diff
@@ -4,15 +4,108 @@ const express = require('express');
...
+  res.render('index');
+});
+app.get('/artist-search', (req, res) => {
+  spotifyApi
+    .searchArtists(req.query.q, { limit: 20, offset: 0 })
+    .then((data) => {
+      console.log('artist-search searchArtists');
+
+      if (req.query.help && req.query.help == 1) {
+        res.send(data.body);
```
ahmedhany82, line 49: I am curious what you are trying to do here :)

app.js
```diff
@@ -4,15 +4,108 @@ const express = require('express');
...
+        let albums = data.body.items.map((album) => {
+          let img;
+          if (album.images && album.images.length > 2) {
+            img = album.images[album.images.length - 2].url;
+          } else if (album.images) {
+            img = album.images[0].url;
+          }
+          return { ...album, img };
+        });
+        res.render('albums', { albums });
```
ahmedhany82, line 80: Another way to do it is to just send `data.body.items` to the `res.render` and in the hbs, pick up the first image from the image array `{{this.name}} ` but I think you know this already. You did it in the `artists-search-results.hbs`

ahmedhany82 (commented):
Hey Marcia, 
Great work as usual! Only added a few tiny comments.
All the best,
Ahmed

app.js
mfndnc: by doing this wherever I am, I can changed the URL and add ?help=1 to it and it will show me the JSON returned by Spotify instead of the nice page.  ;-)

app.js
mfndnc: the first image (images[0].url) was the biggest one. 
Since I did not need that, I was trying to see if a smaller image was available, but not the smallest one

app.js
mfndnc: General question: After you revised the homework, what would happen if I keep working on this fork? Would your comments disappear?

PR #1480, opened 2021-04-18
[Berlin-WDFT-Mar2021] Katka Orlovska 

atkat:


views/artist-search-results.hbs
```diff
@@ -0,0 +1,10 @@
+  {{#each artists}}
+
+    <div >
+        <h2 class="name">{{name}}</h2>
+        <img src="{{images.0.url}}" alt="{{name}}">
```
wax-bear, line 5: There might be no image.
Hence, sometimes no image is displayed.
You could fix this similarly to the way you fixed that issue on line 45 in app.js.

wax-bear (commented):
Great stuff Katka!
Functionally this is all working as expected.
Please check my comments for potential improvements.

PR #1485, opened 2021-04-19
done

neburnodrog:


app.js
```diff
@@ -12,7 +13,62 @@ app.set('views', __dirname + '/views');
...
 // Our routes go here:
 
-app.listen(3000, () => console.log('My Spotify project running on port 3000 🎧 🥁 🎸 🔊'));
+
+app.get('/', (req, res, next) => {
+    res.render('index');
+})
+
+
+app.get('/artist-search/', (req, res, next) => {
```
annawaszak, line 33: I would get rid of the slash after the path to keep it consistent

app.js
```diff
@@ -12,7 +13,62 @@ app.set('views', __dirname + '/views');
...
+
+    const artist = req.query.artist
+
+    spotifyApi
+        .searchArtists(artist, { limit: 6 })
+        .then(data => {
+
+            res.render('artists-results', {
+                artists: data.body.artists.items,
+                data: data,
```
annawaszak, line 43: You don't need to pass the data as you don't make use of it in the 'artists-results' hbs file.

views/tracks.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="container-collection">
+{{#each tracks}}
+    <div class="container-element">
+        <h3>
+            {{name}}
+        </h3>
+        {{#if uri}}
+        <audio src="{{uri}}"></audio>
```
annawaszak, line 8: - Add "controls" attribute to have play/pause button displayed
- Instead of using "uri", you should use "preview_url"

annawaszak (commented):
Hi Ruben,

Great job on the lab! I left a few small improvement suggestions in the comments below. If you have any questions, let me know.

Thanks,
Anna

PR #1486, opened 2021-04-19
BerlinWD03-2021-Rimbert Richter

rimboriri:
done


annawaszak (commented):
Hi Rimbo,

Excellent job! I could not find much to improve here. And use of async await was a nice touch :)

Thanks,
Anna

PR #1507, opened 2021-05-14
[PT-AMS-MAR21] Mohamud Issa

C3lestial-star:
-Viewing the API data in order to structure your code was a bit difficult. I used res.send to print data on the HTML and used that to structure my code. Let me know if anyone knows a better method.

### DIFF
--- a/routes/artist.js
+++ b/routes/artist.js
@@ -2,11 +2,9 @@ require('dotenv').config();
 
 const express = require('express');
 const app = express();
-const hbs = require('hbs');
+const router = express.Router();
 const SpotifyWebApi = require('spotify-web-api-node');
-const album = require('./routes/album');
-const artist = require('./routes/artist');
-const track = require('./routes/track');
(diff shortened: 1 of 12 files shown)


routes/artist.js
```diff
@@ -0,0 +1,36 @@
...
+    clientSecret: process.env.CLIENT_SECRET
+});
+
+// Retrieve an access token
+spotifyApi
+  .clientCredentialsGrant()
+  .then(data => spotifyApi.setAccessToken(data.body['access_token']))
+  .catch(error => console.log('Something went wrong when retrieving an access token', error));
+
+router.get("/artist-search", (req, res) =>{
```
guido-an, line 27: This is a cleaner way to write the same route: https://prnt.sc/132pq7q. 
Same on the next ones.

And just for your knowledge, this would be the way to write the same route using _async_ _await_: https://prnt.sc/132pxw9
You can choose the method you prefer!

PR #1544, opened 2021-06-27
BER - WDFT - May 2021 - Juan Ramirez

Mangonando:


views/tracks.hbs
```diff
@@ -0,0 +1,13 @@
...
+  <div>
+    <h3>{{name}}</h3>
+    <iframe
+      src="https://open.spotify.com/embed?uri={{uri}}"
+      width="300"
+      height="380"
+      frameborder="0"
+      allowtransparency="true"
+      allow="encrypted-media"
+    ></iframe>
```
ahmedhany82, line 11: Another way to do it is like so:
  ```
      <div>
            <audio controls>
            <source src="{{this.preview_url}}" type="audio/mpeg">
            Your browser does not support the audio element.
            </audio>
        </div>
```

ahmedhany82 (commented):
Hey Juan,
Amazing work! I just added a small remark for the tracks part.
Best regards,
Ahmed

PR #1588, opened 2021-09-03
[BER WDFT Aug 2021] Esteban

esteban-p:
All but the styling... 🙈


atkat (commented):
Hey Esteban,

Well done, I really don't have much to say, maybe just that it's good to already get used to removing old console logs after you've finished. 

Let me know if you have any questions about the content. 

Best,
Katka

PR #1637, opened 2021-10-17
[FT-RMT-092021] Jeannette Valencia

JeannetteValencia:
Hi!
This Lab has also been really difficult to me since I cannot put into practice what we learnt in class. I have been reading the instructions several times, but I am afraid of not understanding what I am supposed to do or how to address the exercise. For example, When we code app.get we render the page with the variable that contains the data that will be displayed in the requested page. I am not sure how to do so, or why the displayed data contains all the artists which contains the name introduced.

app.get('/artist-search', (req, res)=>{
  spotifyApi
    .searchArtists(req.query.artist)
    .then(data => {
      console.log('>>>ARTISTS INFO>>>', data.body.artists.items);
      const artistList = data.body.artists.items;
      res.render("artist-search-results", artistList)
    })
    .catch(err => console.log('The error while searching artists occurred: ', err));

})

not sure if I explained this correctly
Regards
J

### DIFF
--- a/views/artist-search-results.hbs
+++ b/views/artist-search-results.hbs
@@ -1,0 +1,10 @@
+<h4>Artist Search Result</h4>
+<div class="artist-data">
+  {{#each this}}
+    <div class="artist-search">
+      <img src="{{images}}">
+      <p>Name: {{name}}</p>
+      <button class="submit-btn" href="/albums/{{uri}}"> View Album</button>
+    </div>
+{{/each}}
+</div>
--- a/views/artist-search-results.hbs
+++ b/views/artist-search-results.hbs
@@ -1,10 +1,10 @@
-<h4>Artist Search Result</h4>
+<h2>Artist Search Result</h2>
 <div class="artist-data">
   {{#each this}}
     <div class="artist-search">
       <img src="{{images}}">
       <p>Name: {{name}}</p>
-      <button class="submit-btn" href="/albums/{{uri}}"> View Album</button>
+      <button class="submit-btn" href="{{id}}"> View Album</button>
     </div>
 {{/each}}
 </div>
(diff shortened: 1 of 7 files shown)


views/index.hbs
```diff
@@ -0,0 +1,8 @@
+<div class"homepage">
+  <form class="search" action="/artist-search" method="GET">
+    <label>
+      <input type ="text" name="artist" size="50"/>
+    </label>
+    <button class="submit-btn" href="http://127.0.0.1:3000/artist-search">Search for an artist</button>
```
TA-Remote, line 6: Never use local links in your code , if you put this code online, this localhost link does not exists there.

In this case it is enough to write href="/artist-search"

views/artist-search-results.hbs
```diff
@@ -0,0 +1,10 @@
+<h4>Artist Search Result</h4>
+<div class="artist-data">
+  {{#each this}}
+    <div class="artist-search">
+      <img src="{{images}}">
+      <p>Name: {{name}}</p>
+      <button class="submit-btn" href="/albums/{{uri}}"> View Album</button>
```
TA-Remote: Here, we want to use the artist id, in order to dispplay all albums.

href="/albums/{{this.id}}"

views/artist-search-results.hbs
```diff
@@ -0,0 +1,10 @@
+<h4>Artist Search Result</h4>
+<div class="artist-data">
+  {{#each this}}
+    <div class="artist-search">
+      <img src="{{images}}">
```
TA-Remote, line 5: If you console.log {{image}}, you will see that in database you have several. images providd - it is an array. So in order to display one in the browser, you need to specify which array elemnt you are using.

TA-Remote (commented):
Hi Jeannette! great to see that you are trying to express what is not clear to you. We can have a look together, try to see and analyze what information are we trying to read (from our database), and how to get it correctly

PR #1639, opened 2021-10-18
FT-RMT-092021] - Alice

AliceXix:
I couldn't remember nor find any solution to paginate that I would fully grasp. Sadly - was really curious about it, hope we see it soon in class again!
I put quite some effort in my CSS after finishing all the iterations as this is not my strong suit and I want to improve it. Would love some feedback/improvement tips on that.

Great start in the weeeeeeeeek 🥳


views/index.hbs
```diff
@@ -1,11 +1,17 @@
-<h1>This is the homepage</h1>
+<div id="container">
+
+    <img id="background-img" src="/images/spotify-background.jpeg">
```
TA-Remote, line 2: At the moment, when I view your app in the browser, the container element is not full screen.

Here I would suggest rather than creating an image tag, just add a background image to your div with id 'container'. Also set that container element to the full width and height of the page, so for example 
```


#container {
  background-image: url("/images/spotify-background.jpeg");
  background-position: center;
  background-size: cover;
   /* position: relative; */
    text-align: center;
    height: 100vh;
    width: 100vw;
}
```

public/styles/style.css
```diff
@@ -117,4 +115,14 @@ body {
...
+  }
+
+  .artist-img {
+      min-width: 320px;
+      min-height: 320px;
+  }
+
+  .album-img {
+      min-width: 300px;
+      min-height: 300px;
```
TA-Remote, line 127: Maybe you have noticed, but your images overlap when you are in artists page (when you search artist)

Here when you set min-width or min-height, keep in mind that if the image provided is bigger than this -> it will be displayed in original size. 

For artists for example, you could set smaller dimensions, and make them fixed -> so they are all the same size.
Or put those images inside a div, and make div of fixed size-so they align nice on the page. and then format img accordingly-to fit inside the div

PR #1676, opened 2021-11-15
 [FTLIS102021] Anne

annemedan:
all iterations working, only styles missing


stefanopisani:
Hey 👋 

In each route there are some extra ".then", make sure you use it only when there is a promise ;) 

Besides this, everything is good, well done!

PR #1677, opened 2021-11-15
LIS-WDFT-FabioGuerreiro

Fabio-FG:
Done until iteration 4


stefanopisani:
All good with your code, nothing to add 🙌🏻
This lab was very long no worries for the last iteration, the important is that you understood how it works ;)

PR #1690, opened 2021-11-16
done, i didnt do the last two myself, because sick, but i know how it works

heiligerhuegel:
… works

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,18 +1,85 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
 
 const app = express();
 
-app.set('view engine', 'hbs');
-app.set('views', __dirname + '/views');
-app.use(express.static(__dirname + '/public'));
+app.set("view engine", "hbs");
+app.set("views", __dirname + "/views");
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -1,18 +1,85 @@
...
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:id", (req, res) => {
+  const id = req.params.id;
+  spotifyApi.getArtistAlbums(id, { limit: 10, offset: 20 }).then(
+    function (data) {
+      console.log("Album information", data.body);
+      res.render("albums", { albums: data.body.items });
```
stefanopisani, line 62: you can save data.body.items in a variable as you did in the previous iteration, just for having a cleaner code ;)

app.js
```diff
@@ -1,18 +1,85 @@
...
+    })
+    .catch((err) =>
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:id", (req, res) => {
+  const id = req.params.id;
+  spotifyApi.getArtistAlbums(id, { limit: 10, offset: 20 }).then(
+    function (data) {
```
stefanopisani, line 60: it works perfectly like this as well but you could try to be more consistent with the syntax

stefanopisani:
Very well done!! keep going 🚀

PR #1692, opened 2021-11-16
solved

joanaCruzWD:
I have a error at the final iteration, and couldn't find it


pmcgarrett:
Good job using the async await functions! Remember to always check if the routes are correct.

PR #1774, opened 2022-02-07
[REMOTE-FT-WD-012022]Tobias Mulatsch

Rospini:


views/albums.hbs
```diff
@@ -4,9 +4,9 @@
     {{#each album}}
```
TA-Remote, line 4: Here,  variable called `album` does not exist.  If you check app.js , inside this route `app.get('/albums/:artistsId'` , currently you have this: `res.render('albums', albumData)`

- If you want to send `albumData`, that is correct, but this variable has no name, so in view you have to refer to it as `this`
- Otherwise you can send `{albumData}` which is equivalent to `{albumData: albumData}  `and then your variables goes by this name, so in view you can refer to it as `albumData`. However this is an object and you want to iterate through an array , which is albumData.albums

PR #1805, opened 2022-04-22
iterations 1 and 2

sina-kaabi:


TA-Remote:
Hi Sina, I wanted to take a look at your lab. 
Question: How did you test your code? After cloning and installing I get errors. One after the other. First, you declared spotifyApi twice. Instead of putting your client secret and client id in app.js, you would put it in .env. Nobody should be able to see this information. Also you miss brackets for the following get request:
```
app.get('/', (req, res, next) => {
    res.render('home');
})
```
After this is fixed, I get an Authetication error. Did you create an account on spotify developers?
There are correct approaches in your code, but overall it can't have worked that way.

TA-Remote:
Please look at the solutions in the Student Portal and try to understand why it works the way it does. If you have any questions we can go over the lab again together.

TA-Remote:
Also: your views folder is inside the public folder.

PR #1812, opened 2022-04-24
WDFT - RUBEN POELEN - IRONSPOTIFY

coderKron:
solved lab with minimal styling. Not in the mood for styling. Never in the mood for styling. Just not getting it the way i want it. Im going to have to practice styling more. But not like this.


TA-Remote:
hahaha I love the green button on your home page though, good job!

PR #1825, opened 2022-04-25
until iteration 4

lukecin:


app.js
```diff
@@ -1,18 +1,52 @@
...
+  );
+
+app.get("/artist-search", (req, res, next) => {
+  spotifyApi
+  .searchArtists(req.query.artist)
+  .then((data) => {
+    console.log(data.body.artists);
+    console.log("The received data from the API: ", data.body);
+    res.render("artist-search-results", data)
+  })
```
TA-Remote, line 39: You do not want to send a massive `data `object into view. Instead you should be sending more specific information - what you tried to console.log actually! 
So for example  `res.render("artist-search-results", data.body.artists.items)`

views/artist-search-results.hbs
```diff
@@ -0,0 +1,3 @@
+<h1>{{data.body.artists}}</h1>
+<img src="{{this}}" alt="artists.image">
+<button><a href="/albums/someArtistIdGoesHere">View Albums</a></button>
```
TA-Remote, line 3: You are sending an array of objects into the view, which means you need a loop here to iterate over it. 

If information that you are sending looks like this:
```res.render("artist-search-results", data.body.artists.items)```

It means it has no name --> so we must use keyword this to access information:

```
{{#each this}}
    <h1>{{this.name}}</h1>
    <img src="{{this.images.2.url}}" alt={{this.name}}>
    <button><a href="/albums/someArtistIdGoesHere">View Albums</a></button>
{{/each}}
```

TA-Remote (commented):
Hey Dani, good start on the lab. Would be amazing if you can find time to make some further progress.

Don't forget we have central support, so you could go analyze the aspects of the lab that are not clear :)

PR #1840, opened 2022-05-07
[WDFT0422] Ana Pinho

AnaGonPinho:
Styling pendent

### DIFF
--- a/views/artist-search-results.hbs
+++ b/views/artist-search-results.hbs
@@ -1,0 +1,11 @@
+
+  {{#each findArtists}}
+  <div>
+    <h1>{{this.name}}</h1>
+    <button><a href ="/albums/{{this.id}}" class="btn btn-primary"></a> View Albums</a>  </button> 
+
+
+     
+    <img src = {{this.images.0.url}} />
+    </div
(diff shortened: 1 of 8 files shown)


views/index.hbs
```diff
@@ -0,0 +1,16 @@
+<html lang="en">
```
webLisbonTA, line 1: you don't need to add the <head> and <body> tag when using handlebars with a layout.
Although you can add a head tag if you need a specific css stylesheet for the view.

views/artist-search-results.hbs
```diff
@@ -0,0 +1,11 @@
+
+  {{#each findArtists}}
+  <div>
+    <h1>{{this.name}}</h1>
+    <button><a href ="/albums/{{this.id}}" class="btn btn-primary"></a> View Albums</a>  </button>
```
webLisbonTA, line 5: should remove the </a> before the label for the button to work properly
`    <button><a href ="/albums/{{this.id}}" class="btn btn-primary"> View Albums</a>  </button> `

webLisbonTA (commented):
Well Done, Ana!

PR #1842, opened 2022-05-07
[FTLIS042022] matildeCosta

matildedacosta:
All done :) I tried to play around with css a bit, used the colors of Spotify! It was a bit hard in some of the pages to make it look always nice, since some searches have more artists then others and then some have more albums and more tracks etc, but I think I got to a middle point! The only hard thing was in the tracks page I couldn't give a gap between the name and the preview or the head row and the other rows below. Tried the flex gap, margins... not sure if I was trying in the wrong place or if there's a different way to do it?


webLisbonTA (commented):
Great job on the lab!!

PR #1853, opened 2022-05-09
[FTLIS042022] Chloé Faurie

chloe4E:
help needed: stuck at iteration 2, can not display result on page

### DIFF
--- a/views/artist-search-results.hbs
+++ b/views/artist-search-results.hbs
@@ -1,0 +1,5 @@
+<div>{{#each beersFromApi}}
+    <h1> {{this.body.name}}</h1>
+
+  {{/each}}
+</div>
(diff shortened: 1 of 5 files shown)


views/artist-search-results.hbs
```diff
@@ -0,0 +1,17 @@
+  <div class="artists">
+      
+  </div>{{#each findArtists}}
```
webLisbonTA, line 3: in the app.js file you are passing the data to the view with the name results, so in the view you need to say {{#each result}}

if you prefer you can rename it in the app.js file to ` res.render("artist-search-results", { findArtists: results });
`

webLisbonTA (commented):
Hi Chloe, take a look on the comments below!

PR #1865, opened 2022-06-07
lab-express-spotify test

zarahsara:


views/home.hbs
```diff
@@ -0,0 +1,6 @@
+<title> Search for an Artist</title>
+
+<form id = "form" role = "search">
```
Matteo-Ironhack, line 3: if you don't use `action` with the form, then the form redirects to the same page. In this case, it should have probably redirected to `/artist-search`

PR #1867, opened 2022-06-07
home page with the form works, and a list of albums gets displayed

nataliabuzila:
the rest of iterations don't work

### DIFF
--- a/views/albums.hbs
+++ b/views/albums.hbs
@@ -1,0 +1,9 @@
+<div>
+    {{#each albums}}
+        <p>{{this.name}}</p>
+        <img src='{{this.image_url}}'/>
+        <a href="/tracks/{{this._id}}">View Tracks</a>
+    {{else}}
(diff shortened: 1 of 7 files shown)


views/artist-search-results.hbs
```diff
@@ -0,0 +1,9 @@
+<div>
+    {{#each artists}}
+        <img src='{{this.image_url}}'/>
+        <p>{{this.name}}</p>
+        <a href="/albums/{{this._id}}">View Albums</a>
```
Matteo-Ironhack, line 5: the api response send back an `id`, not an `_id`. Not every API uses MongoDB

views/albums.hbs
```diff
@@ -0,0 +1,9 @@
+<div>
+    {{#each albums}}
```
Matteo-Ironhack, line 2: the api sends back an `items` array, not an `albums` array

PR #1868, opened 2022-06-07
not enough time for me

EvgeniiaRiaguzova:


keys.env
```diff
@@ -0,0 +1,2 @@
+CLIENT_ID=your clientId goes here
```
Matteo-Ironhack, line 1: this file should just be killed `.env` and not be pushed; it is a security risk

app.js
```diff
@@ -5,14 +5,47 @@ const hbs = require('hbs');
...
+    .clientCredentialsGrant()
+    .then(data => spotifyApi.setAccessToken(data.body['access_token']))
+    .catch(error => console.log('Something went wrong when retrieving an access token', error));
 // Our routes go here:
 
+app.get('/', (req, res) => {
+    res.render('home');
+  });
+
+  spotifyApi
```
Matteo-Ironhack, line 33: writing code like this means that you are not sure you set the access token before trying to get this data. It is one of the traps of asynchronous programming.

PR #1869, opened 2022-06-07
done

ihab5520:


app.js
```diff
@@ -1,18 +1,75 @@
 require('dotenv').config();
 
-const express = require('express');
-const hbs = require('hbs');
+import express, { static } from 'express';
```
Matteo-Ironhack, line 3: you cannot use imports like this in node

app.js
```diff
@@ -1,18 +1,75 @@
...
+console.log('The error while searching artists occurred: ', err)
+);
+} );
+
+app.get('/albums/:artistId', (req, res, next) => {
+spotifyApi
+.getArtistAlbums(req.params.artistId, { limit: 6 })
+.then((data) => {
+res.render('albums', { albums: data.body.items });
+.catch((err) =>
```
Matteo-Ironhack, line 55: this `.catch` is inside the `.then`. It should be after it

app.js
```diff
@@ -1,18 +1,75 @@
...
 
 const app = express();
 
 app.set('view engine', 'hbs');
 app.set('views', __dirname + '/views');
-app.use(express.static(__dirname + '/public'));
+app.use(static(__dirname + '/public'));
 
 // setting the spotify-api goes here:
+const spotifyApi = new SpotifyWebApi({
```
Matteo-Ironhack, line 15: you didn't require the package `spotify-web-api-node`

PR #1886, opened 2022-07-09
WD RT JUN 2022 | Ruben Andino

rubenanlo:
All iterations  plus style of each website. This version is before the super bonus which I will attempt today/tomorrow


TA-Remote:
Amazing Ruben! 🙌

PR #1893, opened 2022-07-11
Nic Solitom - almost done

nicsolitom:
I got stuck with a syntax+typo on the last iteration, but found it only now so I will fix it and style it properly later today.


TA-Remote:
good job Nick,  
You also can add a **`button`** in home page for search input ;)

PR #1894, opened 2022-07-11
Jack Le - incomplete

jackletran:
got stuck in the album iteration, didnt know how to get album id params from url dynamically

### DIFF
--- a/views/albums.hbs
+++ b/views/albums.hbs
@@ -1,0 +1,16 @@
+<h1>Here are the albums!</h1>
+
+{{#each info}}
+  <ul>
+    <li>
+      <a href="/">{{this.name}}</a>
+      <button type="submit">
+        <a href="/albums/{{this.id}}">View Albums</a>
+      </button>
+    </li>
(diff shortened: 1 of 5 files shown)


views/albums.hbs
```diff
@@ -0,0 +1,16 @@
+<h1>Here are the albums!</h1>
+
+{{#each albums}}
+  <ul>
+    <li>
+      <a href="/">{{this.name}}</a>
```
TA-Remote, line 6: you dont need to put the artist name in a tag!  it makes sense when this a tag goes to albums of this artist,not home page :)

views/artist-search-result.hbs
```diff
@@ -0,0 +1,16 @@
+<h1>Here are your results!</h1>
+
+{{#each info}}
+  <ul>
```
TA-Remote, line 4: you can show image of each album here, you need to check if image exist, then show them:
``{{#if this.images.0.url}}
            
            {{/if}}

PR #1897, opened 2022-07-11
Lara Rockenstein // WD FT June22

LMRBLN:


TA-Remote:
You forgot to add .hbs extention to album page,, and because of that you can't see albums list in browser

views/artist-search-results.hbs
```diff
@@ -0,0 +1,16 @@
+<section id="artist-details">
+
+    {{#each everyArtist}}
+    <h1>{{this.name}}</h1>
+
+    {{#if images}}
```
TA-Remote, line 6: here you need to write if this.images instead of images
you are in a loop .

PR #1926, opened 2022-08-02
without css

AdrianaLeitao:


app.js
```diff
@@ -12,7 +13,30 @@ app.set('views', __dirname + '/views');
...
+  
+  // Retrieve an access token
+  spotifyApi
+    .clientCredentialsGrant()
+    .then(data => spotifyApi.setAccessToken(data.body['access_token']))
+    .catch(error => console.log('Something went wrong when retrieving an access token', error));
 
 // Our routes go here:
+app.get('/', (req, res, next) => {
+  res.send(req.query)
```
webLisbonTA, line 29: Here, you had to render the index page instead of sending the query to the frontend

PR #1927, opened 2022-08-02
[FTLIS072022] Brenda Lopes 

Brenda-Lop:


app.js
```diff
@@ -12,7 +12,51 @@ app.set('views', __dirname + '/views');
...
+      res.render('artist-search-results', {artistObj}) 
+    })
+    .catch(err => console.log('The error while searching artists occurred: ', err));
+
+    });
+    
+
+    app.get('/albums/:artistId', (req, res, next) => {
+    spotifyApi    
+    .getArtistAlbums(req.query.button)
```
webLisbonTA, line 51: Here, because it comes from the url, you would need to use "req.params"

webLisbonTA (commented):
Good job on iterations 1 and 2!

PR #1959, opened 2022-09-12
Solved lab

avarap:


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
trebeljahr, line 1: don't commit this ^^

app.js
```diff
@@ -1,18 +1,94 @@
...
+app.get("/albums/:artistId", async (req, res) => {
+
+  const artistResponse = await spotifyApi.getArtist(req.params.artistId)
+  .catch((err) =>
+    console.log("The error while searching artists occurred: ", err)
+  );
+
+  const Sresponse = await spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .catch((err) =>
```
trebeljahr, line 63: you start duplicating this .catch handler -> maybe instead use try catch and have one error handling for this whole block? 

also in doing so you would fix that certain things that are still not properly error handled -> i.e. when Sresponse.body is undefined for example, because the API doesn't return the correct thing for some reason your could would kill the server.

app.js
```diff
@@ -1,18 +1,94 @@
...
+});
+
+app.get("/albums/:artistId", async (req, res) => {
+
+  const artistResponse = await spotifyApi.getArtist(req.params.artistId)
+  .catch((err) =>
+    console.log("The error while searching artists occurred: ", err)
+  );
+
+  const Sresponse = await spotifyApi
```
trebeljahr, line 61: the name Sresponse is weird. First off it starts with a capital letter - suggesting a class - second -> it's a kind of acronym, but not really... just destructure -> and get out items renamed as albums from the body ^^ and you don't have to worry about names haha

app.js
```diff
@@ -1,18 +1,94 @@
...
+
+  const data = { doctitle: artistResponse.body.name + " Albums", albums: albumsItems };
+  res.render("albums",data);
+});
+
+app.get("/tracks/:albumId", async (req, res) => {
+  /* HOW TO SEND SOMETHING ELSE */
+  const Sresponse = await spotifyApi
+    .getAlbumTracks(req.params.albumId)
+    .catch((err) =>
```
trebeljahr, line 81: when I see the same code, spread over a project 3 times, I really start thinking if I should refactor this into a function of some sort ^^

views/tracks.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="container">
+  <div class="row">
+    {{#each tracks}}
+      <div class="tracksCard">
+        <img src="{{this.images.[0].url}}" alt="{{this.name}}" />
```
trebeljahr, line 5: cool -> I didn't even know that [i] mixed in with . notation was a thing in HBS :) today I learned :+1:

trebeljahr (commented):
very good work :) just a few minor things here and there.

PR #1962, opened 2022-09-13
WDPT BER Patrick

Patrick-Groot:


views/albums.hbs
```diff
@@ -0,0 +1,12 @@
+<a href="http://localhost:3000">Back to search</a>
+
+<h1>Albums:</h1>
+<main>
+  {{#each albums}}
+    <div class="block">
+      <h2>{{this.name}}</h2>
+      <img src="{{this.images.0.url}}" alt="Image of {{this.name}}" />
```
trebeljahr, line 8: Image is redundant in alt texts ... most screen readers already announce it anyways ^^

views/index.hbs
```diff
@@ -0,0 +1,4 @@
+<form action="/artist-search" , method="GET">
```
trebeljahr, line 1: probably this should be a POST method...

views/tracks.hbs
```diff
@@ -0,0 +1,12 @@
+<a href="http://localhost:3000">Back to search</a>
```
trebeljahr, line 1: this is not good. The main reason why is because it is specific to your machine and will break on a deployed machine. 
instead use relative urls! i.e. href="/" which would be the homepage, no matter where stuff is hosted. I.e. if your host is mydomain.com -> "/" would still work -> whereas localhost:3000 would not.

app.js
```diff
@@ -1,18 +1,75 @@
...
+      },
+      function (err) {
+        console.error(err);
+      }
+    );
+});
+
+app.get("/tracks/:albumId", (req, res, next) => {  
+  spotifyApi.getAlbumTracks(req.params.albumId).then(
+    function (data) {
```
trebeljahr, line 64: I personall dislike function syntax like this within .then blocks. => arrows are so much shorter and to me at least, more beautiful. 

also, maybe try writing this code with async/await as a practice.

trebeljahr (commented):
good work, a bit low on the styling side but hey :man_shrugging:  
I would probably have done the same ^^

app.js
Patrick-Groot: Yeah, I think I agree. If I remember correctly this is how it was done in the material on IronHack, so I just wanted to try it out once, too :)

Patrick-Groot:
> good work, a bit low on the styling side but hey man_shrugging I would probably have done the same ^^

Thanks for the feedback! :)

PR #1964, opened 2022-09-13
[PTBER072022] berndHavekost

hvkst:
No object destructuring yet, I have to practice more, will try to implement that later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,5 +12,11 @@
   "license": "ISC",
   "devDependencies": {
     "nodemon": "^2.0.2"
+  },
+  "dependencies": {
+    "dotenv": "^16.0.2",
+    "express": "^4.18.1",
+    "hbs": "^4.2.0",
+    "spotify-web-api-node": "^5.0.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "app.js",
   "scripts": {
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -4,13 +4,19 @@
   "description": "",
   "main": "app.js",
   "scripts": {
-    "dev": "nodemon app.js",
+    "dev": "nodemon --ext js,mjs,json,hbs,css app.js",
```
trebeljahr, line 7: ^^

views/artist-search-result.hbs
```diff
@@ -0,0 +1,36 @@
...
+
+  {{! This is not working yet... }}
+  {{!-- <form action='/next-results/{{searchQuery}}/10' , method='GET'>
+    <button id='index-btn' type='submit'>Show more Results</button>
+  </form> --}}
+
+</div>
+
+{{! This is pretty weird... Handlebars doc says I should write the img src like 
+"this.images.[1].url" and that this "this.images.1.url" is wrong
```
trebeljahr, line 34: both are correct actually. 
https://stackoverflow.com/a/13704081

trebeljahr (commented):
nice work. 

This is the second lab where you seemed to be struggling with using async/await vs. or combined with promises, please have another look at it -> if you want we can also go over some more examples or something together during lab time, just let me know on Slack :)

PR #1965, opened 2022-09-13
Spotify lab

jdlms:


app.js
```diff
@@ -1,18 +1,72 @@
-require('dotenv').config();
+test
```
trebeljahr, line 1: ?

app.js
```diff
@@ -1,18 +1,72 @@
...
+
+app.get("/artist-search", async (req, res) => {
+  try {
+    const data = await spotifyApi.searchArtists(req.query.name);
+    res.render("search", {
+      docTitle: "Artist Results",
+      data: data.body.artists.items,
+    });
+  } catch (error) {
+    console.log("The error while searching artists occurred: ", error);
```
trebeljahr, line 42: niceee :) 
the first person to use clean async and await style code so far :)

trebeljahr: makes me happy. 
:roll-safe: doesn't work here :( 
neither does :lel: -saaad.

app.js
```diff
@@ -1,18 +1,72 @@
...
+});
 
-// Our routes go here:
+app.get("/tracks", async (req, res) => {
+  try {
+    const data = await spotifyApi.getAlbumTracks(req.query.id);
+    console.dir(data, { depth: null });
+    res.render("albumtracks", {
+      docTitle: "Tracks",
+      data: data.body.items,
```
trebeljahr, line 65: think about some better names than "data" - maybe "tracks" ? ^^

trebeljahr: and above, "albums" ?

views/searchResults.hbs
```diff
@@ -0,0 +1,14 @@
+<div id="search">
+
+  <p>Albums & Singles by:</p>
```
trebeljahr, line 3: careful with & and other special chars in HTML -> sometimes they break, there is syntax for escaping it though! 
https://stackoverflow.com/questions/7381974/which-characters-need-to-be-escaped-in-html

trebeljahr (commented):
really good job. 
the async/await cleanness made me happy :+1:

PR #1967, opened 2022-09-15
lab done

herrFernandez:
Some styling missing though. Thank you for the hints! :-)

### DIFF
--- a/app.js
+++ b/app.js
@@ -4,7 +4,7 @@ const express = require('express');
 const hbs = require('hbs');
 
 // require spotify-web-api-node package here:
-
+const SpotifyWebApi = require('spotify-web-api-node');
 const app = express();
 
 app.set('view engine', 'hbs');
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -13,6 +13,61 @@ app.use(express.static(__dirname + '/public'));
...
+  });
+
+
+  // search by id (albums)
+
+  app.get('/albums/:artistId', (req, res, next) => {
+
+    spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .then(async (data) => {
```
trebeljahr, line 50: why the async function here? 
this is a promise's .then callback -> it doesn't really need to be an async function + you are not even using await inside the function either, so async makes 0 sense at this place tbh.

app.js
```diff
@@ -13,6 +13,61 @@ app.use(express.static(__dirname + '/public'));
...
+      });
+
+
+  // show tracks of an album
+
+  app.get('/albums/:artistId/:albumId', (req, res) => {
+
+    spotifyApi
+    .getAlbumTracks(req.params.albumId, { limit : 5, offset : 1 })
+    .then(async (data) => {
```
trebeljahr, line 64: same here.

views/artist-search-result.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if artists}}
+    {{#each artists}}
```
trebeljahr, line 2: I am not 100% sure, but I think #each doesn't do anything if there is nothing to #each over... so it should be fine to drop that 1 level of nesting I think.

views/tracks.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="trackrow">
+    {{#if tracks}}
+        {{#each tracks}}
+        <p>{{this.preview_url}}</p>
+            <h1>{{this.artists.[0].name}}</h1>
```
trebeljahr, line 5: this might potentially break, in case that the artist at pos 0 is the "wrong" artist...

trebeljahr: in case of a feature or something like this. better to either -> display all of them, or send the "correct" artist from the server based on the actual artistId used in the search.

app.js
trebeljahr: if you want - try refactoring this whole promise based code in the route handlers -> into async/await code. 
Hint: spotifyApi.getAlbumTracks can be awaited, because it's a promise :) 
and we know it's a promise because we use .then and .catch on it ;)

trebeljahr (commented):
good work. Have one more look at refactoring the promise based code you wrote to async/await for practice though!

PR #2008, opened 2022-10-30
[LISFT202210] Rafael Coelho - Done, restored corrupted version from external hard drive.

RFCoelho-gh:


webLisbonTA:
Good job on the lab Rafael! Next time try to do CSS

PR #2041, opened 2022-11-14
[BER-FTWD-2022] Wei Geng

parrr-geng:


app.js
```diff
@@ -1,18 +1,83 @@
...
+     })
+    .catch((err) =>{console.error(err);})
+    
+})
+
+app.get('/tracks/:albumId', (req, res, next)=>{
+    console.log(req.params.albumId);
+
+    spotifyApi
+    .getAlbumTracks(req.params.albumId, { limit : 10, offset : 1 })
```
matigobbi, line 63: nice that you were able to use  limit and offset Wei!

matigobbi (commented):
nice work Wei, looks very clean and you also used partials ;),

Keep doing !! :D

PR #2082, opened 2022-12-16
WDRMNOV2022 Rumeli Paul

Rumeli33:
Did all the iterations without the css. will work on the css later.


Rumeli33:
Finished lab

Rumeli33:
I found a bug regarding typing invalid artist name, so fixed that.

PR #2162, opened 2023-02-10
[LISFT0123] Erika

erikasoaress:


erikasoaress:
Really difficult and frustrating lab. Could not finish the styling. I feel like we need to do more similar things in class before jumping into something like this.

joaopdg:
Hi Erika, when I cloned your lab the file `package.json` came without the required dependencies to do the lab. This means that probably you forgot to do this step `npm install express hbs spotify-web-api-node dotenv`.

After installing everything your lab looks great, everything is working and you even did the css, congrats!

About the difficulty of the lab, don't worry you will keep working with hbs and express, so there will be plenty of opportunities to keep improving your skills!

PR #2167, opened 2023-02-11
[LISFT0123] Maria Carvalho

miacsilva:


app.js
```diff
@@ -13,6 +14,65 @@ app.use(express.static(__dirname + '/public'));
...
+
+  .catch(err => console.log('The error while searching artists occurred: ', err))
+});
+
+app.get('/tracks/:albumId', (req, res, next) => {
+
+  const id = req.params.albumId 
+
+  spotifyApi
+  .getArtistAlbums(id)
```
webLisbonTA, line 69: Here, you passed the wrong API method.
Instead of `.getArtistAlbums()` that is the method you used above, you should've used the `.getAlbumTracks()` method.
Fixing that solves the problem.

webLisbonTA (commented):
Good job on the lab Maria.

PR #2168, opened 2023-02-11
[LIS0123FT] Paulo Caetano

paulocaetano96:
quite the difficult lab. had issues with the audio tag and how to return the results of our search from the api, after the input had been made. all sorts of confusion here


paulocaetano96:
edit: had issues with <audio> the tag*

lzaquine (commented):
Great job on the lab, Paulo!!

PR #2173, opened 2023-02-12
[LISFT0123] MarishaDeroubaix

Deroubaix:


views/track.hbs
```diff
@@ -0,0 +1,9 @@
+{{#each track}}
+<div >
+  <h1>Title</h1>
```
lzaquine, line 3: Here, instead of Title, you could've passed `{{this.name}}` to display each song's name!

lzaquine (commented):
Great job on the lab, Marisha!!

PR #2179, opened 2023-02-12
[FTLIS0123] Déborah

deborahdelimaa:
some problems with the last iteration and how to link different files to css.

### DIFF
--- a/app.js
+++ b/app.js
@@ -61,7 +61,7 @@ app.get("/albums/:artistId", (req, res, next) => {
     );
 });
 
-app.get("/albums/:trackId", (req, res) => {
+app.get("/tracks/:trackId", (req, res) => {
     spotifyApi.getAlbumTracks(req.params.trackId)
     .then(function(data) {
       console.log(data.body.items);
--- a/app.js
+++ b/app.js
@@ -1,18 +1,77 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -1,18 +1,77 @@
...
+    .then((data) => {
+      console.log("The received data from the API: ", data.body.items[0]);
+      res.render("albums", { result: data.body.items });
+    })
+    .catch((err) =>
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:trackId", (req, res) => {
```
joaopdg, line 64: About the last iteration, if you try to access the tracks of an album you're getting `Cannot GET /tracks/6x5UJEgPpRGdXjpeRY5QSz` 

This is happening because you're sending the user to the path `/tracks/...` but in the `app.js` file you forget to change the path of the route `app.get("/albums/:trackId", (req, res) => {` 

If you change `/albums/:trackId` to `/tracks/:trackId` it will work.

views/layout.hbs
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="/styles/style.css">
```
joaopdg, line 7: About linking different `css` files. 
You can link each file inside the `hbs` page that you want it to be applied, instead of linking it in the `layout.hbs`.

joaopdg (commented):
Great job on the lab Déborah!

PR #2183, opened 2023-02-13
[LISFT0123] Francisco Pereira

FranciscoManuelPereira:
- Iteration 5: I had difficulty doing this exercise, I couldn't find the right path in the terminal...
I did it with the help of a colleague

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,18 +1,98 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
+const SpotifyWebApi = require("spotify-web-api-node");
 
 const app = express();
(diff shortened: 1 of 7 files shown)


app.js
```diff
@@ -1,18 +1,98 @@
...
+    .then((data) => {
+      res.render("albums", { albums: data.body.items });
+      //console.log("Album information", data.body);
+    })
+    .catch((err) =>
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
```
webLisbonTA, line 72: Here, you can't have the same root. One `id` should refer to the `artist`, and the other `id` refers to the `albums`, in order for you to get the `tracks`. Otherwise, the code will be stuck in the first `get `method you have here

webLisbonTA (commented):
Good job on the lab Francisco.

PR #2185, opened 2023-02-13
[LISFT0123] Gabriel

gbrlcsgn:


app.js
```diff
@@ -1,18 +1,65 @@
...
+      let allArtists = data.body.artists.items;
+      res.render("artist-search-results", { allArtists });
+    })
+    .catch((err) =>
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
+  spotifyApi.getArtistAlbums().then(
```
webLisbonTA, line 49: You need to do the `object destructuring`, in order to pass an argument to the `API method` you are using.
After that, after the promise is resolved, you will get the `data`, with the information you want. Then, you need to dig deeper into the `data` in order to filter the information that you want to send for the frontend. 
When you have the info that you want, you can store it inside a `variable`, and when rendering your `view`, pass the `variable` with the filtered data, in form of an `object`.
```
app.get('/albums/:artistId', (req, res, next) => {
  let { artistId } = req.params;
  spotifyApi.getArtistAlbums(artistId).then(
    (data) => {
      console.log('Artist albums', data.body.items);
      let allAlbums = data.body.items;
      res.render('albums', { allAlbums });
    },
    function (err) {
      console.error(err);
    }
  );
});
```

webLisbonTA (commented):
Good job on the lab. Check the comments for a better understanding.

PR #2186, opened 2023-02-13
[LISWD0123]FARID

Farid-97:


Views/albums.hbs
```diff
@@ -0,0 +1,7 @@
+{{#each
+  artists
```
webLisbonTA, line 2: In your route, you passed the information as `albums` and you are trying to loop over each `artists` to render the information. Just need to change for `albums` and you'll get the data displayed.

PR #2188, opened 2023-02-13
Eveline Coenegrachts

evelinecoen:


views/artist-search-results.hbs
```diff
@@ -0,0 +1,6 @@
+{{#each allArtists}}
+<div class=‘image’>
+  <p>{{this.name}}</p>
+  <a href="/albums/someArtistIdGoesHere">View Albums</a>
```
joaopdg, line 4: Here you can not pass the ID hard coded you need to make it dynamic, the same way you did for the name:

```
<a href="/albums/{{this.id}}">View Albums</a>
```

app.js
```diff
@@ -4,15 +4,58 @@ const express = require('express');
...
+    console.log('The received data from the API: ', data.body.artists.items); // where do we get this (?)
+    // ----> 'HERE'S WHAT WE WANT TO DO AFTER RECEIVING THE DATA FROM THE API'
+    const allArtists = data.body.artists.items   // you have to create a variable to display it as an object
+    res.render('artist-search-results', {allArtists})   // this is when we are going to display the content on the screen
+  })
+  .catch(err => console.log('The error while searching artists occurred: ', err));
+
+})
+
+app.get('/albums/:artistId', (req, res, next) => {
```
joaopdg, line 52: You did the route for the artist search, the route for the albums will follow the same logic.

But now we no longer have a `form` so instead of getting the `input` of the `user` through the `req.query` we will get the `id` of the artist through the `req.params`.

After this you need to make the call to the `API` with the specific method `.getArtistAlbums()` passing the `artistId` inside the `()` .

Wait for the response, and render the `hbs` page with the `response` from the `API`.

joaopdg (commented):
Hi Eveline, try to finish the lab later this week so you can practice a bit more. Overall, good job so far!

PR #2224, opened 2023-03-20
[FTRMT022023] Hugo

hugowannabedev:
Setting the initial route was very confusing at the beginning. And the 'tracks' route is still not working.


TA-Remote:
Hey Hugo!!!! great you played around displaying other properties. Guess you enjoyed working with music related lab😉
Anyway, you just miss a  " / " before 'tracks...  ` app.get('tracks/:albumId', (req, res, next) => { `    and everything is working perfectly🙌

PR #2285, opened 2023-05-06
Spotify API

Renatopais:


views/albums.hbs
```diff
@@ -0,0 +1,26 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Albums</title>
+</head>
+<body>
+    <h1>Albums</h1>
+    {{#each response.items}}
+        <div>
+            <a href="/albums/{{../artistId}}/{{id}}/tracks">View tracks</a>
```
joaopdg, line 14: You're sending the user to the current page `albums` instead of sending to the `tracks` page, also you're passing a lot of info on the route you just need to pass the `id`:
```
 <a href="/tracks/{{id}}">View Tracks</a>
```

joaopdg (commented):
Good job on the lab Renato, check the corrections that I left you below and the lab's solution to have an idea of what you could have done differently!

PR #2287, opened 2023-05-07
[lis-wdft-2023] - Lucas Telles

lltelles:


views/index.hbs
```diff
@@ -0,0 +1,17 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="/styles/style.css"/>
+    <title>Home</title>
+  </head>
+  <body>
+    <div class="btn">
+      <form action="/artist-search" method="GET">
+        <input type="text" name="/artist-search"/>
+        <button type="submit">Search for artist</button>
```
joaopdg, line 13: You already have the html boiler plate code on your `layout.hbs` so here you just need to put the code referring to the `body` of your page.

You should not use the route path as the name of the input:
```
    <form action="/artist-search" method="GET">
      <input type="text" name="search" />
      <button type="submit">Search for an Artist</button>
    </form>
```

PR #2291, opened 2023-05-07
[LIS-FT-APR23] Guy_Dagan

Guy-Dagan:


app.js
```diff
@@ -1,18 +1,69 @@
...
+  .then((data) => spotifyApi.setAccessToken(data.body["access_token"]))
+  .catch((error) =>
+    console.log("Something went wrong when retrieving an access token", error)
+  );
 
 // Our routes go here:
 
-app.listen(3000, () => console.log('My Spotify project running on port 3000 🎧 🥁 🎸 🔊'));
+app.get("/", (req, res) => {
+  res.render("index.hbs");
```
brunorocha20, line 32: You don't need to put the termination of your file since you already declared above that the `view engine` is going to be `hbs`, so, you should only render `res.render("index");`

brunorocha20 (commented):
Good job on the lab. Check the lab solutions for a better understanding.

PR #2293, opened 2023-05-07
[WDFT LIS APR23] Victoria Avelar

avelarvicky:


app.js
```diff
@@ -1,18 +1,78 @@
...
+});
+
+app.get("/albums/:artistId", async (req, res, next) => {
+  try {
+    let albums = await spotifyApi.getArtistAlbums(req.params.artistId);
+
+    console.log("Artist albums: ", albums.body.items);
+
+    //res.render("albums", { results: albums.body.items });
+    res.send({ results: albums.body.items });
```
joaopdg, line 56: You're sending the API response to the view, instead of rendering a .hbs file displaying the API response. You should use the commented line that you have above `res.render( ...` instead of `res.send( ...`

views/album-tracks.hbs
```diff
@@ -0,0 +1,5 @@
+{{#each results}}
+    <img src="{{images.0.url}}" alt="" />
+    <h2>{{name}}</h2>
+    <a href="/albums/{{id}}">view albums</a>
```
joaopdg, line 4: Instead of a button sending you back, you should have a player for the sample of the song

joaopdg (commented):
Good job on the lab Vitoria, check the corrections that I left you below and the lab's solution to have an idea of what you could have done differently!

PR #2294, opened 2023-05-08
Lis-WDFT-2023-MarisaPinheiro

Marisa-Pinheiro:
lab done.Error connecting to API for the track page and css file unable to connect on the albums and track page. please, indicate mistakes. code previously compared to classmates

### DIFF
--- a/views/albums.hbs
+++ b/views/albums.hbs
@@ -1,0 +1,9 @@
+ <body id="albums">
+ {{#each response}} 
+  <div class="albums-div-info">
+  <img class="albums-image" src="{{this.images.0.url}}" alt="{{this.name}}" />
+    <h2 class="albums-name">{{this.name}}</h2>
+    <a href="/albums/:artistId">View Tracks</a>
+    </div>
+  {{/each}}
+</body>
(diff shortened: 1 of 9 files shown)


views/albums.hbs
```diff
@@ -0,0 +1,9 @@
+ <body id="albums">
+ {{#each response}} 
+  <div class="albums-div-info">
+  <img class="albums-image" src="{{this.images.0.url}}" alt="{{this.name}}" />
+    <h2 class="albums-name">{{this.name}}</h2>
+    <a href="/albums/:artistId">View Tracks</a>
```
joaopdg, line 6: You're sending the user to the `albums` page again instead of sending to `tracks` also you have `:artistId` that syntax is for the route in the `app.js` file, here in the `hbs` file you should pass the actual `id`:
```
<a href="/tracks/{{id}}">View Tracks</a>
```

joaopdg (commented):
Good job on the lab Marisa, check the corrections that I left you below and the lab's solution to have an idea of what you could have done differently!

PR #2295, opened 2023-05-08
done

ana-rak:
fifth exercise is not finished

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,18 +1,68 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
+const SpotifyWebApi = require("spotify-web-api-node");
 
 const app = express();
 
-app.set('view engine', 'hbs');
-app.set('views', __dirname + '/views');
-app.use(express.static(__dirname + '/public'));
+app.set("view engine", "hbs");
+app.set("views", __dirname + "/views");
+app.use(express.static(__dirname + "/public"));
 
 // setting the spotify-api goes here:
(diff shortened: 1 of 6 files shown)


app.js
```diff
@@ -1,18 +1,68 @@
...
+    let responseAlbums = await spotifyApi.getArtistAlbums(req.params.artistId);
+    let allAlbums = responseAlbums.body.items;
+    //console.log(allAlbums[0].images);
+    res.render("albums", { allAlbums });
+  } catch (error) {
+    console.error(error);
+  }
+});
+
+app.get("/viewtracks/:albumId", async (req, res)=> {
```
brunorocha20, line 59: In this `app.get` method, you could replicate the logic built for the artist albums, where you received from the `params`, the `id` that you need to make the call to the API. You should save it in a variable, that is going to be the information you'll be sending to your `views`.

views/albums.hbs
```diff
@@ -0,0 +1,6 @@
+<div>
+  {{#each allAlbums}}
+    <img src={{this.images.0.url}} alt="" width="100px" height="100px" />
+  {{/each}}
+  <a href="/albums/someArtistIdGoesHere">View Albums</a>
```
brunorocha20, line 5: This `<a> tag` should be inside the `#each allAlbums`, you needed to receive the `id` of the album in your URL, to gather it on your `req.params` and make the call to the API. 
```
{{#each allAlbums}}
    <img src={{this.images.0.url}} alt="" width="100px" height="100px" />
    <a href="/viewtracks/{{this.id}}">View Tracks</a>
{{/each}}  
```

brunorocha20 (commented):
Good job on the lab. Check the notes and then the solution for a better understanding.

PR #2298, opened 2023-05-08
[LIS-WD-FT-APRIL23] Guilherme Dantas

Gui-Dantas:


app.js
```diff
@@ -12,7 +13,41 @@ app.set('views', __dirname + '/views');
...
+        res.render('artist', {results:data.body.artists.items});
+    } 
+    catch (error) {
+        console.log('The error while searching artists occurred: ', error)
+    } 
+});
+
+app.get('/albums/:artistId', async (req, res, next) => {
+    try{
+       let data = await spotifyApi.getArtistAlbums(req.params.artist)
```
brunorocha20, line 45: Since you wrote that the `URL` to this `route` is `'/albums/:artistId'`, you need to link the params with your route for this call to the `API`  to work properly. 
```let data = await spotifyApi.getArtistAlbums(req.params.artistId)```

brunorocha20 (commented):
Guilherme, check the lab solutions for a better understanding of this lab.

PR #2326, opened 2023-06-04
[FTRMT052023] JuanLozano

Jualolo:


Jualolo:
I think the https://developer.spotify.com/ website stopped working, or maybe I'm doing something wrong but I cannot access the data. On the other hand I feel really lost with this lab :(

PR #2329, opened 2023-06-04
Got up to iteration 4. Struggling with accessing the images. Getting confused with the concept of cleaning up data from response, and passing it hbs

jani500:


jani500:
Got up to iteration 4. Struggling with accessing the images. Getting confused with the concept of cleaning up data from response, and passing it to hbs

PR #2362, opened 2023-07-08
[PT] Franklin

franklinosei:


app.js
```diff
@@ -1,18 +1,63 @@
...
 // setting the spotify-api goes here:
 
 // Our routes go here:
+app.get("/", (req, res) => {
+  res.render("homepage");
+});
+
+app.get("/artist-search", (req, res) => {
+  spotifyApi
+    .searchArtists(req.query.search)
```
iPoolito, line 37: A good practice and for easy to read its to save the querys in variables first and then pass it something like 
```
const searchArtist =req.query.search
  spotifyApi
    .searchArtists( searchArtist)
```

app.js
```diff
@@ -1,18 +1,63 @@
...
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:artistId", (req, res) => {
+  // .getArtistAlbums() code goes here
+  spotifyApi
+    .getArtistAlbums(req.params.artistId)
+    .then((data) => {
+      // console.log(data.body.items);
```
iPoolito, line 55: also try to remove the console.logs everytime you create a PR its for good practice :P you can keep the ones inside catch

iPoolito:
Nice job Franklin also using Partials 🔥

PR #2364, opened 2023-07-10
Done

HadrienMv:


app.js
```diff
@@ -4,15 +4,72 @@ const express = require('express');
...
 // Our routes go here:
+app.get('/home', (rq, rs) => {
+    try {
+        rs.render('homepage')
+    }
+    catch (err) {console.log(err)}
+})
+
+app.get('/artist-search', (rq,rs) => {
+    spotifyApi.searchArtists(rq.query.artist)
```
iPoolito, line 39: A good practice and for easy to read its to save the querys in variables first and then pass it something like
```
const searchArtist =req.query.search
  spotifyApi
    .searchArtists( searchArtist)
    ```

File: app.js [hunk 2]
```diff
@@ -4,15 +4,72 @@ const express = require('express');
...
+    .then(data => {
+        // console.log('The received data from the API: ', data.body.artists.items[0]);
```
iPoolito, line 41: also try to remove the console.logs everytime you create a PR its for good practice 🔥  you can keep the ones inside catch

iPoolito:
Nice work Hadrien 🚀 also as an advise, remember that we can use the layout in HBS? as the name said you only need to create a file with the name of layout.hbs and could be something like this
```
<html lang="en">

<head>
    <meta charset="UTF-8" />
    <title>Document</title>
</head>

<body>
    {{{body}}}
</body>

</html>
```

PR #2367, opened 2023-07-10
[PT] Biljana

BiljanaKotev:


app.js
```diff
@@ -1,18 +1,84 @@
...
+    })
+    .catch((err) => {
+      console.log(err);
+    });
+});
 
 // Our routes go here:
 
-app.listen(3000, () => console.log('My Spotify project running on port 3000 🎧 🥁 🎸 🔊'));
+app.get("/homepage", (req, res) => {
```
iPoolito, line 41: In general and by convention every time that we reffer to homepage the route its  '/'

app.js
```diff
@@ -1,18 +1,84 @@
...
+app.set("view engine", "hbs");
+app.set("views", __dirname + "/views");
+app.use(express.static(__dirname + "/public"));
 
 // setting the spotify-api goes here:
+app.get("/", (req, res, next) => {
+  spotifyApi
+    .searchArtists("madonna")
+    .then((data) => {
+      // console.log(data.body);
```
iPoolito, line 31: Also when you dont need any more this logs try to remove it every time you create a PR 🚀

app.js
```diff
@@ -1,18 +1,84 @@
...
 // Our routes go here:
 
-app.listen(3000, () => console.log('My Spotify project running on port 3000 🎧 🥁 🎸 🔊'));
+app.get("/homepage", (req, res) => {
+  res.render("homepage");
+});
+
+app.get("/artist-search", (req, res) => {
+  spotifyApi
+    .searchArtists(req.query.artistName)
```
iPoolito, line 47: also another way to do it its first create a variable for example
`const artistName=req.query.artistName`
and put the variable instead the full path, it would be like 
```
spotifyApi
    .searchArtists( artistName)
```

views/albums.hbs
```diff
@@ -0,0 +1,24 @@
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Document</title>
+  </head>
+  <body>
```
iPoolito, line 7: Because we already have our layout its not necessary to write the tags for html, meta, and body because our layout are actually already doing it for us 🔥

iPoolito:
Excelente Work Biljana 🚀  Also good work with styles , i leave you some comments just to keep it in mind 😎

app.js
BiljanaKotev: noted and thansk!

app.js
BiljanaKotev: Noted

app.js
BiljanaKotev: Noted and thanks!

BiljanaKotev:
Thanks Pool!

PR #2368, opened 2023-07-10
[PT] Maria Garcia

MariaG6:


app.js
```diff
@@ -2,17 +2,63 @@ require('dotenv').config();
...
+  .catch(error => console.log('Something went wrong when retrieving an access token', error));
 
 // Our routes go here:
+app.get('/', (req, res) => {
+  res.render('homepage')
+})
+
+app.get('/artist-search', (req, res) => {
+  spotifyApi
+  .searchArtists(req.query.artist)
```
iPoolito, line 31: A good practice and for easy to read its to save the querys in variables first and then pass it something like
```
const searchArtist =req.query.artist
  spotifyApi
    .searchArtists( searchArtist)
    ```

File: app.js [hunk 2]
```diff
@@ -2,17 +2,63 @@ require('dotenv').config();
...
+})
+
+app.get('/albums/:artistId', (req, res) => {
+  const id = req.params
+  console.log(id.artistId)
+  spotifyApi
+  .getArtistAlbums(id.artistId)
+  .then( data => {
+    const albums = data.body.items
+    console.log(albums);
```
iPoolito, line 47: Also just for keep code clean and good practices when you create a new Pull request try to remove the console.logs that you dont need, for example the ones that are inside the catch you can leave it

views/albums.hbs
```diff
@@ -0,0 +1,23 @@
+<html lang="en">
```
iPoolito, line 1: Because we already create a layout its not necessary to add again all the html,body and meta tags because our layout is already handle that for us 😎

PR #2369, opened 2023-07-11
final lab

TiagoLobo-source:


app.js
```diff
@@ -1,18 +1,74 @@
...
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
 
 // require spotify-web-api-node package here:
+const SpotifyWebApi = require("spotify-web-api-node");
+console.log(process.env.CLIENT_ID);
```
iPoolito, line 8: Its cool try to know if the things are working, but every time we create a Pull Request we should remove the console.logs that are not necessary, you can leave the ones that are inside catch

app.js
```diff
@@ -1,18 +1,74 @@
...
 
 // setting the spotify-api goes here:
 
-// Our routes go here:
+app.get("/", (req, res) => {
+  res.render("home");
+});
+app.get("/artist-search", (req, res) => {
+  spotifyApi
+    .searchArtists(req.query.artist)
```
iPoolito, line 35: A good practice and for easy to read its to save the querys in variables first and then pass it something like
```
const searchArtist =req.query.artist
  spotifyApi
    .searchArtists( searchArtist)
```

iPoolito:
Nice work Tiago 🚀 also as an advise, remember that we can use the layout in HBS? as the name said you only need to create a file with the name of layout.hbs and could be something like this
```

<html lang="en">

<head>
    <meta charset="UTF-8" />
    <title>Document</title>
</head>

<body>
    {{{body}}}
</body>

</html>
```

PR #2370, opened 2023-07-11
[PT] Xavier Ortega

Xavieros03:


app.js
```diff
@@ -4,15 +4,76 @@ const express = require('express');
...
+const spotifyApi = new SpotifyWebApi({
+  clientId: process.env.CLIENT_ID,
+  clientSecret: process.env.CLIENT_SECRET,
+});
+
+// Retrieve an access token
+spotifyApi
+  .clientCredentialsGrant()
+  .then((data) => {
+    console.log({body: data.body})
```
iPoolito, line 18: Its cool try to know if the things are working, but every time we create a Pull Request we should remove the console.logs that are not necessary, you can leave the ones that are inside catch

app.js
```diff
@@ -4,15 +4,76 @@ const express = require('express');
...
 
-// setting the spotify-api goes here:
 
 // Our routes go here:
+app.get("/", (req, res) => {
+  res.render("index");
+});
+
+app.get("/artist-search", (req, res) => {
+  const artist = req.query.artist;
```
iPoolito, line 38: nice job! 🔥

views/albums.hbs
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
```
iPoolito, line 1: Because we already create a layout its not necessary to add again all the html,body and meta tags because our layout is already handle that for us 😎

PR #2371, opened 2023-07-11
[PT]Bright

brytemenz:


app.js
```diff
@@ -1,18 +1,80 @@
-require('dotenv').config();
+require("dotenv").config();
 
-const express = require('express');
-const hbs = require('hbs');
+const express = require("express");
+const hbs = require("hbs");
```
iPoolito, line 5: Its a good practice put all your imports at the top of the file 

```
`const app = express();

app.set("view engine", "hbs");
app.set("views", __dirname + "/views");
app.use(express.static(__dirname + "/public"));`
```

app.js
```diff
@@ -1,18 +1,80 @@
...
 // Our routes go here:
+app.get("/", (req, res) => {
+  res.render("index");
+});
+
+app.get("/artist-search", (req, res) => {
+  spotifyApi
+    .searchArtists(req.query.ArtistName)
+    .then((data) => {
+      console.log(data.body.artists.items[0]);
```
iPoolito, line 39: Another advise its that everytime that you deliver a PR try to remove all the logs that are not need it, for example the ones that are inside a catch, you can leave it 🚀

app.js
```diff
@@ -1,18 +1,80 @@
...
+    .catch((err) =>
+      console.log("The error while searching artists occurred: ", err)
+    );
+});
+
+app.get("/albums/:artistId", (req, res, next) => {
+  const artistId = req.params.artistId;
+
+  spotifyApi
+    .getArtistAlbums(artistId)
```
iPoolito, line 51: this is excelente 🔥  first saving the params in a variable and then pass it like this, is way better and easy to read

iPoolito:
Nice work Bright 🚀  also as an advise, remember that we can use the layout in HBS? as the name said you only need to create a file with the name of  layout.hbs  and could be something like this 

```
<html lang="en">

<head>
    <meta charset="UTF-8" />
    <title>Document</title>
</head>

<body>
    {{{body}}}
</body>

</html>
```

iPoolito:
So you dont need to write in each hbs file again the meta tags and the body and html

iPoolito:
Keep an eye also for the CSS and styling

PR #2374, opened 2023-07-13
[PT] Marcos Francesquini

marcosfrancesquini:


app.js
```diff
@@ -12,7 +23,58 @@ app.set('views', __dirname + '/views');
 app.use(express.static(__dirname + '/public'));
 
 // setting the spotify-api goes here:
+app.get("/", (req, res) => {
+    res.render("index")
+})
+//res.render(data.body.artists.items[0].images)
+app.get("/artist-search", (req, res) => {
+    console.log("Artist came from browser - request")
+    console.log(req.query.artist)
```
iPoolito, line 32: Just for good practice every time you create a PR try to remove the console.logs that are not needed, for example you can leave the ones inside the catch

app.js
```diff
@@ -12,7 +23,58 @@ app.set('views', __dirname + '/views');
...
+        .then((data) => {
+
+            const artistImage = data.body.artists.items[0].images[0].url;
+            const artistName = data.body.artists.items[0].name;
+            const artistId = data.body.artists.items[0].id;
+            // console.log(artistObject)
+            res.render("artist-search-result", { artistImage: artistImage, artistName: artistName, artistId: artistId })
+        })
+
+        //res.render("artist-search-result",{artist:data.body.artists.items})})
```
iPoolito, line 48: Also try to avoid leave commented code that you dont need :D

app.js
```diff
@@ -12,7 +23,58 @@ app.set('views', __dirname + '/views');
...
+        .catch(err => console.log(err))
+})
+
+
+
+app.get("/albums/:artistId/", (req, res, next) => {
+    // .getArtistAlbums() code goes here
+
```
iPoolito, line 56: A good practice and for easy to read its to save the params  in variables first and then pass it something like
```
const artistId =req.params.artistId
  spotifyApi
    .getArtistAlbums( artistId )
    ```

PR #2394, opened 2023-08-04
[LS-FT-JULY23]Gabriel

Ogbera-Gabriel:


brunorocha20 (commented):
Good job on the lab Gabriel. When working with the `hbs`, remember that you only need to use the HTML tags and file formation once. In the `layout.hbs` you do that, everywhere else you just use the tags and `hbs` syntax inside.

PR #2399, opened 2023-08-06
[LISFTJULY23] Julia Guillotte

jsguillotte:


jacobviv (commented):
Good work with your lab Julia. When working with the hbs, remember that you only need to use the HTML tags and file formation once. In the layout.hbs you do that, everywhere else you just use the tags and hbs syntax inside.

PR #2402, opened 2023-08-06
[LIS-FT-JULY23] Bruno Cunha

brcunha3000:


brunorocha20 (commented):
Good job on the lab Bruno. Just remember that when working inside the hbs `#each` loop, sometimes it works without referencing upwards, but it's better practices if you always use the `this.` keyword to reference the object you are looping over.

PR #2408, opened 2023-08-07
[LIS-FT-072023] João Elias

johnseliseto:


brunorocha20 (commented):
Good job on the lab João. When working with `hbs`, remember that you only need to use the HTML tags and file formation once. In the `layout.hbs` you do that, everywhere else you just use the tags and `hbs` syntax inside.

PR #2410, opened 2023-08-07
LIS-FT-JULY23] Mariana Marques

mfilipa97:


jacobviv (commented):
Good job with your lab Mariana. When working with the hbs, remember that you only need to use the HTML tags and file formation once. In the layout.hbs you do that, everywhere else you just use the tags and hbs syntax inside.

PR #2413, opened 2023-08-07
[FT-LIS-André Pinto]

pinto-andre:


brunorocha20 (commented):
Good job on the lab André. When working with `hbs`, remember that you only need to use the HTML tags and file formation once. In the `layout.hbs` you do that, everywhere else you just use the tags and `hbs` syntax inside.

pinto-andre:
> 
Right, I was forgetting about that. Thanks!

